PR #4, opened 2019-05-07
dashboard: use kube-system namespace

dkoshkin:
This was incorrectly changed when moving configs to this repo. Reverting back so the pod works.


dkoshkin:
Closing in favor of https://github.com/mesosphere/kubeaddons-configs/pull/5

PR #6, opened 2019-05-08
change prometheus retention to 7d

s12chung:
Justin Barrick reports 10d to be pretty high,
due to prometheus resource consumption


s12chung:
after further discussion, we agreed to leave the defaults and close this.

PR #12, opened 2019-05-09
Add kubernetes version options to Addon CRD

shaneutt:


crds/kubeaddons_v1alpha1_addon.yaml
```diff
@@ -20,6 +20,10 @@ spec:
       properties:
         spec:
           properties:
+            kubernetes:
+              properties:
+                minversion: string
```
hectorj2f: `minVersion` or `minSupportedVersion`, same for `maxVersion`. We then should validate that as part of the Cluster Configuration CRD 🤔, we'll need to get the kubernetes installed version

shaneutt: :+1:

PR #16, opened 2019-05-15
elasticsearch: increase settings

hectorj2f:
We are checking if these new settings would tackle recent issues on soak

### DIFF
--- a/templates/elasticsearch.yaml
+++ b/templates/elasticsearch.yaml
@@ -28,7 +28,7 @@ spec:
         resources:
           # need more cpu upon initialization, therefore burstable class
           limits:
-            cpu: 1000m
+            cpu: 2000m
             memory: 2048Mi
           requests:
             cpu: 250m
@@ -39,12 +39,12 @@ spec:
         hooks:
           drain:
             enabled: false
-        heapSize: 1536m
+        heapSize: 3072m
         resources:
           # need more cpu upon initialization, therefore burstable class
           limits:
-            cpu: 1000m
-            memory: 4096Mi
+            cpu: 4000m
+            memory: 8192Mi
           requests:
             cpu: 250m
-            memory: 2048Mi
+            memory: 8000Mi
--- a/templates/elasticsearch.yaml
+++ b/templates/elasticsearch.yaml
@@ -31,7 +31,7 @@ spec:
             cpu: 2000m
             memory: 2048Mi
           requests:
-            cpu: 250m
+            cpu: 500m
             memory: 1536Mi
       data:
         updateStrategy:
@@ -46,5 +46,5 @@ spec:
             cpu: 4000m
             memory: 8192Mi
           requests:
-            cpu: 250m
-            memory: 8000Mi
+            cpu: 1000m
+            memory: 4608Mi


templates/elasticsearch.yaml
```diff
@@ -25,18 +25,18 @@ spec:
...
           # need more cpu upon initialization, therefore burstable class
           limits:
-            cpu: 1000m
-            memory: 4096Mi
+            cpu: 4000m
+            memory: 8192Mi
           requests:
             cpu: 250m
-            memory: 2048Mi
+            memory: 8000Mi
```
jieyu: I feel we should simply disallow bursting for memory. Memory bursting relies on OOM score adjustment, which is not very reliable. I would rather prefer a more deterministic behavior on this.

i.e., change this to 8192Mi

templates/elasticsearch.yaml
```diff
@@ -25,18 +25,18 @@ spec:
...
+        heapSize: 3072m
         resources:
           # need more cpu upon initialization, therefore burstable class
           limits:
-            cpu: 1000m
-            memory: 4096Mi
+            cpu: 4000m
+            memory: 8192Mi
           requests:
             cpu: 250m
```
jieyu: Consider change requests to 1000m

templates/elasticsearch.yaml
```diff
@@ -25,18 +25,18 @@ spec:
         resources:
           # need more cpu upon initialization, therefore burstable class
           limits:
-            cpu: 1000m
+            cpu: 2000m
             memory: 2048Mi
           requests:
             cpu: 250m
```
jieyu: consider change request to 500m

templates/elasticsearch.yaml
```diff
@@ -25,18 +25,18 @@ spec:
...
             memory: 1536Mi
```
jieyu: See below. I'd prefer not bursting memory. i.e., set it to 2048Mi

templates/elasticsearch.yaml
hectorj2f: Disallowing bursting goes against the settings in our ES installations for Kubernetes which recommend
```
In terms of Memory resources you should make sure that you follow that equation:

${role}HeapSize < ${role}MemoryRequests < ${role}MemoryLimits
```

hectorj2f: For instance in: https://github.com/helm/charts/tree/master/stable/elasticsearch

templates/elasticsearch.yaml
hectorj2f: same as above: https://github.com/mesosphere/kubeaddons-configs/pull/16#discussion_r285500317

hectorj2f:
@jieyu Could you review again this PR and what you think about my last comments ?

templates/elasticsearch.yaml
shaneutt: `8000` seems very high when our heapsize is `3072`, could we not adjust our request down to `4096` to significantly improve the pods ability to be scheduled?

Also it does seem like we would want to keep bursting available, as that will provide memory wiggle-room while still giving us a better baseline to get scheduled.

jieyu: OK, that's fair. I remember that we saw memory usage spike to 7G+ in soak. So the limit has to be 8G. The request can be lower.

I think it's OK to reduce memory request to 4608 (1.5x the heap size) given the heap size is 3072.

Thoughts?

templates/elasticsearch.yaml
jieyu: OK. that's fair! Thanks for the pointer.

templates/elasticsearch.yaml
shaneutt: Sounds good :+1:

hectorj2f: Done

hectorj2f: IMHO, we won't see an impact until we start using it, and see if that value is enough or not

hectorj2f:
Done

hectorj2f:
@jieyu @shaneutt I've addressed your comments, please re-review :)

PR #17, opened 2019-05-16
elasticsearch: disabled data drain and update strategy

hectorj2f:
We should consider if we want this disabled by default even when using local storage, or change it to be a configurable value.

### DIFF
--- a/templates/elasticsearch.yaml
+++ b/templates/elasticsearch.yaml
@@ -21,6 +21,8 @@ spec:
             cpu: 100m
             memory: 1536Mi
       master:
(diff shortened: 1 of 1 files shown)


templates/elasticsearch.yaml
```diff
@@ -21,6 +21,8 @@ spec:
             cpu: 100m
             memory: 1536Mi
       master:
+        updateStrategy:
```
jieyu: Please do the same for `data` too.

hectorj2f: Done

PR #20, opened 2019-05-17
Add metrics-server addon

dkoshkin:
This enables HPA and `kubectl top` and dashboards in the kubernetes-dashboard.

### DIFF
--- a/templates/metricsserver.yaml
+++ b/templates/metricsserver.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: metricsserver
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  namespace: kube-system
+  chartReference:
+    chart: stable/metrics-server
+    values: |
+      ---
+      args:
+      # enable this if you have self-signed certificates, see: https://github.com/kubernetes-incubator/metrics-server
+      - --kubelet-insecure-tls


jieyu:
Can we hold-off on this PR? I feel it's a bit redundant with the current monitoring pipeline (based on Prometheus). I just don't know if this is gonna introduce tech debt or not. I need to do more research on this.

dkoshkin:
Yep lets talk about it.

hectorj2f:
@dkoshkin @jieyu I did some research already and it is common to have the metrics-server and prometheus running together to extract additional metrics. That is the benefit of running the metrics server alongside with Prometheus.

However, I found this adapter which replaces the necessity of the metrics-server https://github.com/DirectXMan12/k8s-prometheus-adapter. WDYT ?

hectorj2f:
This adapter can be installed as helm chart https://github.com/DirectXMan12/k8s-prometheus-adapter. It is relevant to have the additional metrics that the metrics-server or adapter collect because those would help us with the performance workload test.

shaneutt:
Marked this as paused because it seems there are some additional considerations in the comments to go over before this is fully ready for review.

jieyu:
@hectorj2f I prefer `k8s-prometheus-adapter`

jieyu:
I am wondering what metrics are not yet in the metrics pipeline that metrics server provides?

hectorj2f:
as said metrics-server is used in other components, but let's give a try to the k8s-prometheus-adapter.

@jieyu mentioned that it is maintained by Google people

shaneutt:
Since we seem to be going with another implementation for the moment, seems like we should close this for now and we can re-open this if we need to revisit.

PR #21, opened 2019-05-19
prometheus: add service monitor for traefik

hectorj2f:


hectorj2f:
```
kubeaddons   traefik-monitor                                      104s
```
```
prometheus-kubeaddons   	1       	Mon May 20 20:12:15 2019	DEPLOYED	prometheus-operator-5.7.0 	0.29.0                      	kubeaddons
```

templates/prometheus.yaml
```diff
@@ -11,7 +11,19 @@ spec:
     values: |
       ---
       prometheus:
+        additionalServiceMonitors:
+          - name: traefik-monitor
+            selector:
+              matchLabels:
+                release: traefik-kubeaddons
```
jieyu: It's weird to have traefik addon specific settings in prometheus addon. What if traefik is not installed (e.g., disabled by the user)?

Is there a way to make this `ServiceMonitor` kind of more general. For instance, it applies to all services under `kubeaddons` namespace with some special label which we can add to those addons from which we want their metrics. For instance, something like the following:

```yaml
additionalServiceMonitors:
  - name: kubeaddons-service-monitor
    selector:
      matchLabels:
        kubeaddons.mesosphere.io/servicemonitor: "true"
    namespaceSelector:
      matchNames:
        - kubeaddons
    endpoints:
      - port: metrics
        interval: 30s
        path: /metrics
```

That means we need to tweak addons (Elastic and Traefik) to expose metrics at the same path with some additional labels on the service. I believe this might be doable.

hectorj2f:
I just talked with Joe, we might change this implementation and try scraping the metrics endpoint for the services in kubeaddons with a certain label

templates/prometheus.yaml
```diff
@@ -11,7 +11,19 @@ spec:
...
+            namespaceSelector:
+              matchNames:
+                - kubeaddons
+            endpoints:
+              - port: metrics
+                interval: 10s
```
jieyu: Please change to 30s for consistency.

hectorj2f:
Closing this PR cause we merged a different approach in #27

PR #22, opened 2019-05-20
calico: use default namespace

dkoshkin:
Use the default namespace.

Fixes https://jira.mesosphere.com/browse/DCOS-54029, requires additional changes in `kubeaddons`.


shaneutt:
LGTM @dkoshkin , however needs signed commits

PR #23, opened 2019-05-20
helm: pin chart versions

dkoshkin:
Fixes https://jira.mesosphere.com/browse/DCOS-53706

Added a new filed in the CRD `version` and then just grabbed the latest versions for the charts we are using.

### DIFF
--- a/crds/kubeaddons_v1alpha1_addon.yaml
+++ b/crds/kubeaddons_v1alpha1_addon.yaml
@@ -30,6 +30,8 @@ spec:
               properties:
                 release:
                   type: string
+                version:
+                  type: string
                 repo:
                   type: string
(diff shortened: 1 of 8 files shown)


alejandroEsc:
I dont understand. Are we just declaring what versions we are using? or is this new field something that can/should be changed? Is the ticket asking that we declare or make it configurable?

dkoshkin:
>We should introduce a mechanism to pin the version fo the helm chat that we're using for each release.

Yeah the ticket only talks about pinning the versions as without that running `helm install` will just install the latest version.

alejandroEsc:
Discussed on channel. This pins the version on Helm install to ensure we use the right chart version.

PR #26, opened 2019-05-21
elasticsearch: add exporter for prometheus

hectorj2f:
Add exporter for elasticsearch

PR #27, opened 2019-05-21
[prometheus] namespace service monitor

hectorj2f:
As discussed with Jie, we will add a service monitor that would aim to collect the info from the kubeaddons namespace and for a specific label. Therefore any service exposing a metrics endpoint and with a label `kubeaddons.mesosphere.io/servicemonitor: "true"` should be scrapped.

### DIFF
--- a/templates/elasticsearchexporter.yaml
+++ b/templates/elasticsearchexporter.yaml
@@ -9,8 +9,15 @@ spec:
     minSupportedVersion: v1.14.0
   chartReference:
     chart: stable/elasticsearch-exporter
+    version: 1.2.0
     values: |
       ---
+      es:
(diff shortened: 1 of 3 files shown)


templates/elasticsearchexporter.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
```
hectorj2f: that is part of another PR #26 but I am testing it here too.

alejandroEsc: Is this something that we should remove later?

hectorj2f: no, #26 would go before #27

PR #29, opened 2019-05-23
Add Velero

shaneutt:
This PR adds Velero to the manifest based addons.

### DIFF
--- a/templates/velero.yaml
+++ b/templates/velero.yaml
@@ -1,0 +1,378 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: velero
+  namespace: kubeaddons
+spec:
+  manifest: |
+    ---
+    apiVersion: v1
+    kind: List
+    metadata:
+      name: velero-crds
+      namespace: velero
+    items:
+    - apiVersion: apiextensions.k8s.io/v1beta1
+      kind: CustomResourceDefinition
+      metadata:
+        creationTimestamp: null
+        labels:
+          component: velero
+        name: schedules.velero.io
+      spec:
+        group: velero.io
+        names:
+          kind: Schedule
+          plural: schedules
+        scope: Namespaced
+        version: v1
+    - apiVersion: apiextensions.k8s.io/v1beta1
+      kind: CustomResourceDefinition
+      metadata:
+        creationTimestamp: null
+        labels:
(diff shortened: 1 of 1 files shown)


s12chung:
i'm not sure if we should move away from helm for this....? just to make it work for the new kubeaddons.

shaneutt:
@s12chung part of the intent here was to package minio and velero together into a single Addon, given what's provided in the velero repository as the helm chart doesn't provide options to deploy minio alongside it (though velero provides an example minio setup which we're now using).

joejulian:
I feel charts are the units we should be focused on supporting. This tech debt we're banking for charts that don't do what we need should be fixed by working upstream on improving our adding charts.

We could fix Velero by adding Minio as a subchart and installing them both together.

shaneutt:
> We are loosing all the benefits of using helm values yaml definitions cry. Why don't we deploy both together via kubeaddons as done for elasticsearch + elasticsearchexporter ?

I'm not confident in the benefits of using helm charts vs. manifest based implementations @hectorj2f. I personally feel like this way of packaging Velero provides a bit more clarity about components, at the cost of some verbosity.

> I'd suggest to have a separate template for minio.

A separate addon?

> This is a nightmare to maintain YAML in YAML.

This is a problem we already had, and we have it whether we use manifests or helm values. We need to solve this problem, but I would contend that this PR is not the place to do so, and simply having _less_ yaml in yaml by using helm charts is not significantly helpful I think.

Let me know your thoughts.

shaneutt:
@hectorj2f & @joejulian upstream chart improvements, subcharts, and other things sound nice, and I'm not opposed, but could we agree that this is a separate concern from what's trying to be accomplished here and would it be a fair compromise to look at this problem more wholesale later (as per a JIRA that we can spin up as a follow up to this)? It seems like we're currently discussing on unspoken rules, I think having a conversation and follow up that leads to actual rules would be more appropriate, let me know your thoughts.

shaneutt:
Closing this, will revisit later

hectorj2f:
@shaneutt What did it change since my latest review ?

hectorj2f:
I thought we were having a pending discussion about this PR, but it seems we don't have it.

shaneutt:
To continue discussion on this, the changes were pulled and https://github.com/mesosphere/kubeaddons-configs/pull/31 has been created as a follow up.

PR #30, opened 2019-05-23
Improve Traefik Fault Tolerance

shaneutt:
Increases the number of default traefik pods deployed to help reduce ingress downtime from individual node failures.

### DIFF
--- a/templates/traefik.yaml
+++ b/templates/traefik.yaml
@@ -11,7 +11,7 @@ spec:
     version: 1.68.4
     values: |
       ---
-      replicas: 3
+      replicas: 2
       service:
         labels:
           kubeaddons.mesosphere.io/servicemonitor: "true"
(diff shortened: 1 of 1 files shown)


s12chung:
can you write why we need more replicas? a card or something?

templates/traefik.yaml
```diff
@@ -11,6 +11,7 @@ spec:
     version: 1.68.4
     values: |
       ---
+      replicas: 3
```
hectorj2f: two replicas should be enough. Did we face any specific issue to increase the replicas ?

shaneutt: hang on, this is `wip` I'm looking at the potential autoscaling feature and will come back to this shortly.

shaneutt:
Closing for the moment to get it off the queue, will re-open later

templates/traefik.yaml
hectorj2f: Alright :), roger that.

shaneutt:
@hectorj2f ready for re-review, skipping autoscale for the moment and focusing on a simple and quick turnaround for the moment.

PR #31, opened 2019-05-24
Add Velero

shaneutt:
The purpose of this PR is to add Velero, and convert it from the old way we use to deploy it. This PR works to cover the following issues:

* Deal with flakiness in Minio deployments
* Isolate Velero Minio so that's its solely used by Velero
* Add persistent storage to Minio for backups
* Reduce custom logic for Velero, and move remaining logic to extraSteps
* Enable daily cluster-wide backups
* Converts the last few remaining workers so that we can fully deprecate the old style workers (after helm)

At this time its specifically being added using the manifest implementation to better couple together the related minio service and velero quickly and expediently, but [DCOS-54304](https://jira.mesosphere.com/browse/DCOS-54304) was created as a follow up to this to utilize helm and perhaps make upstream contributions to the charts to assist with our use case.

### DIFF
--- a/templates/velero.yaml
+++ b/templates/velero.yaml
@@ -1,0 +1,389 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: velero
+  namespace: kubeaddons
+spec:
+  manifest: |
+    ---
+    apiVersion: v1
+    kind: List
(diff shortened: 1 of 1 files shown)


shaneutt:
Decided in standup that we're going to go with this implementation for the time being, and we have several follow up JIRAs for later iterations on this.

PR #33, opened 2019-05-27
prometheus-adapter: add adapter in replacement of metrics-server

hectorj2f:
As an alternative to #20, it has been discussed whether we want to deploy metrics-server with the overlaps that it includes when having prometheus-operator. Therefore there is an alternative https://github.com/DirectXMan12/k8s-prometheus-adapter. That seems to replace metrics-server and is maintained by RH and Google people.


hectorj2f:
As discussed with Jie in Slack, we'll give a try to this chart.

PR #35, opened 2019-05-30
Add velero resources

shaneutt:
Adds some baseline resource configurations to Velero and its Minio counterpart.

### DIFF
--- a/templates/velero.yaml
+++ b/templates/velero.yaml
@@ -225,6 +225,16 @@ spec:
                 mountPath: "/storage"
               - name: config
                 mountPath: "/config"
+              resources:
+                # these resources are currently kept relatively low, as this minio
+                # instance is ONLY being used by Velero at this time. Feel free
+                # to change these based on new metric information from test clusters.
+                requests:
+                  memory: 256Mi
+                  cpu: 250m
+                limits:
+                  memory: 512Mi
+                  cpu: 500m
     - apiVersion: v1
       kind: Service
       metadata:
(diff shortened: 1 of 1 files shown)


templates/velero.yaml
```diff
@@ -358,7 +368,15 @@ spec:
               ports:
               - containerPort: 8085
                 name: metrics
-              resources: {}
+              resources:
+                # by default velero uses around 1.5Mi while idle, so we give 50Mi of breathing room for scheduling
+                # due to the nature of the work Velero does, and a desire to minimize disruptions to it, we provide relatively high resource limits.
+                requests:
```
hectorj2f: Could we compare these values with what we get from soak, so we can know how much it normally consumes  ?

Also, we should trigger a backup and check what are the new consumption when it is actively working. 

That is how we estimated some resource usage utilization.

s12chung:
as a note, we might move this addon back to a helm chart because the `velero` addon isn't idempotent atm: https://jira.mesosphere.com/browse/DCOS-54642

templates/velero.yaml
shaneutt: > Could we compare these values with what we get from soak, so we can know how much it normally consumes?

Yes we could do this, however its important to note that these values were derived from the values provided in the previous helm implementation and so I would say such work is outside of the scope for this iteration, as this change is simply getting us back to something [close to] the setup we had previously.

>Also, we should trigger a backup and check what are the new consumption when it is actively working.
> That is how we estimated some resource usage utilization.

Yes we should do this. I've created hDCOS-54782 as a follow up for this.

I propose we continue with these settings for the time being, and follow up with DCOS-54782. If you would prefer to not merge this as-is, my intention will be to close this for now in deference to DCOS-54782 as this work was done as a quick side-fix, but if the scope must increase then we need to vet that as part of our regular time budgets.

hectorj2f: My concern are the limits, would 1k millicores and 1G handle a backup of 50 nodes cluster ? What about a 20 nodes cluster ? That is why we could kind of estimate based on the usage for few nodes ? 

If we define limits and then the pod exceeds them, the backup will fail because the pod got killed due to an OOM.  We can specify the requested resources.

shaneutt: Understood thanks for the update @hectorj2f :+1: I'm going to defer this for now.

shaneutt:
Deferred in favor of DCOS-54782

PR #49, opened 2019-06-05
setup fluentbit-prometheus properly

hectorj2f:
it wasn't pointing to the correct endpoint. 

The endpoint in fluent-bit is different than `/metrics` that is why we have to define the specific path `/api/metrics/prometheus`.

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -21,7 +21,17 @@ spec:
               matchNames:
                 - kubeaddons
             endpoints:
-              - port: metrics
+              - path: metrics
+                interval: 30s
+          - name: fluentbit-service-monitor
+            selector:
+              matchLabels:
+                kubeaddons.mesosphere.io/fluentbitservicemonitor: "true"
+            namespaceSelector:
+              matchNames:
+                - kubeaddons
+            endpoints:
+              - path: api/v1/metrics/prometheus
                 interval: 30s
         prometheusSpec:
           enableAdminAPI: true
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -13,20 +13,20 @@ spec:
       ---
       prometheus:
         additionalServiceMonitors:
-          - name: kubeaddons-service-monitor
+          - name: kubeaddons-service-monitor-metrics
             selector:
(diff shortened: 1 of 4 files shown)


templates/prometheus.yaml
```diff
@@ -21,7 +21,17 @@ spec:
               matchNames:
                 - kubeaddons
             endpoints:
-              - port: metrics
+              - path: metrics
```
hectorj2f: this has to be port

templates/prometheus.yaml
```diff
@@ -21,7 +21,17 @@ spec:
...
+                interval: 30s
+          - name: fluentbit-service-monitor
+            selector:
+              matchLabels:
+                kubeaddons.mesosphere.io/fluentbitservicemonitor: "true"
+            namespaceSelector:
+              matchNames:
+                - kubeaddons
+            endpoints:
+              - path: api/v1/metrics/prometheus
```
hectorj2f: add port metrics here too

templates/prometheus.yaml
```diff
@@ -23,6 +23,17 @@ spec:
             endpoints:
               - port: metrics
                 interval: 30s
+          - name: fluentbit-service-monitor
```
jieyu: Let's not do this. Please rely on `kubeaddons-service-monitor`

templates/prometheus.yaml
```diff
@@ -26,7 +26,7 @@ spec:
           - name: fluentbit-service-monitor
```
jieyu: Can you change the name `kubeaddons-service-monitor-api-v1-metrics-prometheus

templates/prometheus.yaml
```diff
@@ -16,7 +16,7 @@ spec:
           - name: kubeaddons-service-monitor
```
jieyu: Change this to `kubeaddons-service-monitor-metrics`

templates/prometheus.yaml
s12chung: done.

templates/prometheus.yaml
s12chung: done

PR #50, opened 2019-06-05
bump up grafana cpu

s12chung:
seeing prometheus warning

Screen Shot 2019-06-05 at 3 05 53 PM

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -72,7 +72,7 @@ spec:
         resources:
           # keep request = limit to keep this container in guaranteed class
           limits:
-            cpu: 100m
+            cpu: 200m
             memory: 100Mi
           requests:
             cpu: 100m


templates/prometheus.yaml
```diff
@@ -72,10 +72,10 @@ spec:
         resources:
           # keep request = limit to keep this container in guaranteed class
           limits:
-            cpu: 100m
+            cpu: 200m
             memory: 100Mi
           requests:
-            cpu: 100m
+            cpu: 200m
```
hectorj2f: `100m`

hectorj2f: Screen Shot 2019-06-05 at 11 40 49 PM

hectorj2f: this is the Cpu since the last time we deployed grafana... let's check if tomorrow we get a bigger spike when having more dashboards in grafana. But I assume we can remain with the current settings.

did you reproduce this alert again ? is it only happening when creating the cluster ? That could be negligible

s12chung: it happened twice that's all.

s12chung: we can try tomorro.w

s12chung: changed though.

PR #51, opened 2019-06-06
rename phoenix to konvoy

s12chung:


s12chung:
it's the id. the logo are embedded into the html as an svg, which hasn't changed.

s12chung:
still need to change the image asked umr about it.

s12chung:
it is........ i admit, i double checked diffs as a sanity check.

PR #56, opened 2019-06-10
prometheus rules for etcd

alejandroEsc:
adding our ow rules to prometheus, here we remove the false positive alerts.

### DIFF
--- a/prometheus-rules/etcd-rules.yaml
+++ b/prometheus-rules/etcd-rules.yaml
@@ -12,106 +12,93 @@ spec:
   groups:
   - name: etcd
     rules:
+    # lost quorum
     - alert: etcdInsufficientMembers
       annotations:
-        message: 'etcd cluster "{{ $labels.job }}": insufficient members ({{ $value
-          }}).'
-      expr: sum(up{job=~".*etcd.*"} == bool 1) by (job) < ((count(up{job=~".*etcd.*"})
-        by (job) + 1) / 2)
+        message: 'etcd cluster "{{ $labels.job }}": insufficient members ({{ $value}}).'
+      expr: sum(up{job=~".*etcd.*"} == bool 1) by (job) < ((count(up{job=~".*etcd.*"}) by (job) + 1) / 2)
       for: 3m
       labels:
         severity: critical
+    # cluster has not been able to elect a leader
     - alert: etcdNoLeader
       annotations:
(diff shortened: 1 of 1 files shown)


alejandroEsc:
@s12chung sure there you go, not sure what is wrong with copy paste as we need these rules mostly the same?

s12chung:
nothing wrong :). just noting it.

alejandroEsc:
@s12chung double checking you ok with the dir name? @shaneutt can you weigh in also? Just want to make sure we are aligned.

s12chung:
i'm ok with it :)

prometheus-rules/etcd-rules.yaml
```diff
@@ -0,0 +1,117 @@
+# Rules for alerting against etcd, without the etcd alerts: etcdHighNumberOfFailedGRPCRequests
```
shaneutt: comments++

prometheus-rules/etcd-rules.yaml
```diff
@@ -0,0 +1,117 @@
...
+    release: prometheus-kubeaddons
+  name: prometheus-kubeaddons-prom-etcd
+spec:
+  groups:
+  - name: etcd
+    rules:
+    - alert: etcdInsufficientMembers
+      annotations:
+        message: 'etcd cluster "{{ $labels.job }}": insufficient members ({{ $value
+          }}).'
```
shaneutt: Are we trying to be strict about whitespace here? I feel this can just as well be a single line? :thinking:

prometheus-rules/etcd-rules.yaml
```diff
@@ -0,0 +1,117 @@
...
+      annotations:
+        message: '{{ $value }}% of requests for {{ $labels.method }} failed on etcd
+          instance {{ $labels.instance }}.'
+      expr: |-
+        sum(rate(etcd_http_failed_total{job=~".*etcd.*", code!="404"}[5m])) BY (method) / sum(rate(etcd_http_received_total{job=~".*etcd.*"}[5m]))
+        BY (method) > 0.05
+      for: 10m
+      labels:
+        severity: critical
+    - alert: etcdHTTPRequestsSlow
```
shaneutt: How would you feel about a brief comment before each alert, which provides a summary of the alert? At least for ones where we think we can expand upon what the alert name and message indicate?

PR #60, opened 2019-06-20
Add NS for kommander, remove servicemonitor

shaneutt:


templates/kommander.yaml
```diff
@@ -6,6 +6,15 @@ metadata:
...
 spec:
   manifest: |
+    ---
+    apiVersion: v1
+    kind: Namespace
+    metadata:
+      name: kommander
+    spec:
+      finalizers:
+      - kubernetes
```
DanielMSchmidt: What does this do?

shaneutt: Ensures that cleanup occurs when the namespace is deleted

PR #61, opened 2019-06-20
calico: use host-local and pod subnet cidr

hectorj2f:
This instructs calico to use the defined pod subnet cidr instead of defining a hardcoded one.

Note: I am doing some additional tests for validation

### DIFF
--- a/templates/calico.yaml
+++ b/templates/calico.yaml
@@ -34,10 +34,11 @@ spec:
               "type": "calico",
               "log_level": "info",
               "datastore_type": "kubernetes",
-              "nodename": "__KUBERNETES_NODE_NAME__",
+              "node_name": "__KUBERNETES_NODE_NAME__",
               "mtu": __CNI_MTU__,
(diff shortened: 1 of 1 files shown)


hectorj2f:
I am currently facing this issue on AWS https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/214#issuecomment-464098149 ...

hectorj2f:
It seems required to define the value of https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/calico.yaml#L607 accordingly to the podSubnet defined in cluster.yaml. Otherwise AWS EBS CSI pods will fail as well as other pods when trying to communicate to the outside world. I initially tried to comment the value of this env var and instruct calico node to use the pod cidrs, however that triggered the error aforementioned.

hectorj2f:
I am closing this PR cause. We will deploy calico as part of ansible in the future instead as an addon via this hardcoded templates.

PR #62, opened 2019-06-25
prometheus: specify a volume name

hectorj2f:
By setting the name, the auto-generated name via the prometheus-operator will result in `db-prometheus-prometheus-kubeaddons-prom-prometheus-0` which contains 53chars.

```
E0625 04:15:15.056075       1 volume.go:128] Internal Server Error: failed to create volume 'pvc-d480ba57-96ff-11e9-9cea-0a44df461fe2': response: Service "pvc-d480ba57-96ff-11e9-9cea-0a44df461fe2-ctrl-svc" is invalid: metadata.labels: Invalid value: "prometheus-prometheus-kubeaddons-prom-prometheus-db-prometheus-prometheus-kubeaddons-prom-prometheus-0": must be no more than 63 characters
```

```
db-prometheus-prometheus-kubeaddons-prom-prometheus-0   Bound    pvc-66873662-30d8-4f1d-a9f9-c7863b97d7b3   50Gi       RWO            ebs-csi-driver   3m3s
```

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -41,6 +41,8 @@ spec:
           externalUrl: "/ops/portal/prometheus"
           storageSpec:
             volumeClaimTemplate:
+              metadata:
+                name: db
               spec:
                 accessModes: ["ReadWriteOnce"]
                 # 50Gi is the default size for the chart


jdef:
while this solves the immediate problem (of exceeding a 65 char limit?), it looks like a pretty severe length restriction for a volume name at this point because "database" generates a 59-char vol name: expanding an 8-char label to a 59-char label that has a 65-char length limit seems like kicking the can down the road, but not very far. in other words, i think it's likely that we'll run into this again sooner than later.

can we do better than a 59-char label in this case?

there's tons of stutter in the 59-char metric that's generated

jdef:
cc @nfnt

hectorj2f:
@jdef certainly we can by changing the helm chart of the Prometheus operator. I just wanted to solve this immediate problem.

hectorj2f:
@jdef it seems this was already reported but not action was taken https://github.com/coreos/prometheus-operator/pull/2016/files

PR #63, opened 2019-06-26
dex

mhrabovcin:


templates/dex-k8s-authenticator.yaml
```diff
@@ -0,0 +1,40 @@
...
+kind: Addon
+metadata:
+  name: dex-k8s-authenticator
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: raw.githubusercontent.com/dex-k8s-authenticator
+    # TODO(mh): If this is going to be used figure out the production URL
```
shaneutt: If its agreeable for you @mhrabovcin, I would like to have a JIRA created for this TODO so that we can better keep track of this need (currently, we're not allowed to utilize github issues for kubeaddons work organization). Does that sound :heavy_check_mark: ?

templates/dex.yaml
```diff
@@ -0,0 +1,64 @@
...
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: stable/dex
+    version: 1.4.0
+    values: |
+      ---
+      image: mhrabovcin/dex
+      imageTag: "d2f240c9c72682913eaff4fd85d119d7d2d0f240"
```
shaneutt: Might be a nice cordiality to add a comment above this with the entire URL pointing to this revision, just makes it easy for someone to click while viewing the code. What do you think?

templates/dex.yaml
```diff
@@ -0,0 +1,64 @@
...
+          tlsCert: /etc/dex/tls/https/server/tls.crt
+          tlsKey: /etc/dex/tls/https/server/tls.key
+        grpc:
+          addr: 0.0.0.0:5000
+          tlsCert: /etc/dex/tls/grpc/server/tls.crt
+          tlsKey: /etc/dex/tls/grpc/server/tls.key
+          tlsClientCA: /etc/dex/tls/grpc/ca/tls.crt
+        oauth2:
+          skipApprovalScreen: true
+
```
shaneutt: nit: whitespace
```suggestion
```

templates/dex.yaml
```diff
@@ -0,0 +1,64 @@
...
+        staticClients:
+        # This `id` must by in sync with `dex-k8s-authenticator.yaml` value as well as
```
shaneutt: comments++ :+1:

templates/traefik.yaml
```diff
@@ -36,6 +36,10 @@ spec:
         enabled: true
         enforced: true
         generateTLS: true
+        # dex service is exposed with TLS certificate signed by self signed root
+        # Dex CA certificate.
+        # TODO(mh): Figure out if we can configure CA certs per endpoint/ingress
```
shaneutt: Is this another place it would be good to have a JIRA created for follow up (and linked here in the comments)?

templates/dex.yaml
```diff
@@ -0,0 +1,64 @@
...
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: stable/dex
+    version: 1.4.0
+    values: |
+      ---
+      image: mhrabovcin/dex
```
hectorj2f: shouldn't we move this docker image under mesosphere ?

templates/dex.yaml
hectorj2f: can we get a version tag even if it is temporal ?

templates/dex.yaml
```diff
@@ -0,0 +1,64 @@
...
+          tlsClientCA: /etc/dex/tls/grpc/ca/tls.crt
+        oauth2:
+          skipApprovalScreen: true
+
+        staticClients:
+        # This `id` must by in sync with `dex-k8s-authenticator.yaml` value as well as
+        # kube-apiserver flag `oidc-client-id`.
+        - id: kube-apiserver
+          redirectURIs:
+            - 'https://dex-k8s-authenticator-kubeaddons.kubeaddons.svc.cluster.local:5555/dex-k8s-authenticator/callback/kubernetes-cluster'
```
hectorj2f: Note that `cluster.local` could vary although we are not setting it at this moment.

templates/dex-k8s-authenticator.yaml
```diff
@@ -0,0 +1,40 @@
...
+        #tlsKey: /path/to/dex-client.key
+        clusters:
+        - name: kubernetes-cluster
+          short_description: "Kubernetes cluster"
+          description: "Kubernetes cluster authenticator"
+          # client_secret: value is generated automatically - see `setupDexK8sAuthenticator()`
+          # This `client_id` must by in sync with `dex.yaml` value as well as
+          # kube-apiserver flag `oidc-client-id`.
+          client_id: kube-apiserver
+          issuer: https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex
```
hectorj2f: This is an example where this settings cannot work without the dex settings defined below.

templates/dex.yaml
shaneutt: Indeed, it would be nice to just tag anything we're going to use for this purpose for the time being.

templates/dex.yaml
mhrabovcin: I'll add the following comment:

```
            # This URI is just a placeholder and it will be replaced during the
            # deployment with URL pointing to the traefik ingress public
            # load balancer.
```

templates/dex-k8s-authenticator.yaml
mhrabovcin: What would be a preferred way to enforce dependency between `dex` and `dex-k8s-authenticator`?

templates/dex.yaml
mhrabovcin: I am going to propose changes to upstream `dex` repository and ideally, we'll be running off official container. If we won't get changes upstream in the time I am planning to push custom version to our `mesosphere` namespace.

templates/traefik.yaml
mhrabovcin: I've created https://jira.mesosphere.com/browse/DCOS-56033

templates/dex.yaml
shaneutt: As this is high prio and it appears a suitable answer was provided for the concern, I'm going to mark this as resolved but please reopen and flag request changes if this is not the case.

templates/dex-k8s-authenticator.yaml
shaneutt: I believe for the moment and in the interest of time we should skip on worrying about this dependency management issue and allow a manual approach in the interim. I have created [https://jira.mesosphere.com/browse/DCOS-56045](DCOS-56045) to cover the next iteration on this wherein we need to automate this.

templates/dex-k8s-authenticator.yaml
```diff
@@ -0,0 +1,43 @@
...
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: raw.githubusercontent.com/dex-k8s-authenticator
+    # TODO(mh): If this is going to be used figure out the production URL
+    # See: https://github.com/mintel/dex-k8s-authenticator/issues/70
+    repo: https://raw.githubusercontent.com/mintel/dex-k8s-authenticator/feat-helm-repo/charts/repo/stable
+    version: 1.0.0
```
hectorj2f: Shouldn't the version be `version: 1.1.1` as in mesosphere/charts ?

hectorj2f: yes, it has to be that version https://github.com/mesosphere/charts/blob/master/docs/staging/index.yaml#L21

mhrabovcin: Yes, I am still working on an update to this PR where I'll point dex-k8s-authenticator to our custom chart repo and as well move use dex container from `mesosphere/dex` repo.

hectorj2f: 👍

templates/traefik.yaml
```diff
@@ -39,6 +39,12 @@ spec:
         enabled: true
         enforced: true
         generateTLS: true
+        # dex service is exposed with TLS certificate signed by self signed root
+        # Dex CA certificate. It is not clear if traefik supports configuring
+        # trusted certificates per backend. This should be investiaged in a
+        # separate issue.
+        # See: https://jira.mesosphere.com/browse/DCOS-56033
+        insecureSkipVerify: true
```
hectorj2f: the default is `insecureSkipVerify: false` that will affect other usages of traefik :/. 
I don't like that but let's move forward and address it in a different PR. Perhaps we have to set this value for traefik when Dex is enabled.

PR #72, opened 2019-07-03
Add cloudProvider mapping

joejulian:
The cloudProvider field will enable or disable this addon per supported
cloud provider (currently: aws, docker, or none).

Not specifying cloudProvider, or leaving it empty gives the same results
and using the special `all` provider like: `all: enabled`.

### DIFF
--- a/templates/awsstorageclass.yaml
+++ b/templates/awsstorageclass.yaml
@@ -9,6 +9,8 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    aws: disabled
   manifest: |
     ---
     apiVersion: storage.k8s.io/v1
--- a/templates/dex-k8s-authenticator.yaml
+++ b/templates/dex-k8s-authenticator.yaml
@@ -6,6 +6,8 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    all: disabled
   chartReference:
     chart: dex-k8s-authenticator
     repo: https://mesosphere.github.io/charts/staging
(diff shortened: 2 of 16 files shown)


templates/localstorageclass.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
```
s12chung: think it's safe to delete this addon.

templates/localstorageclassdefault.yaml
```diff
@@ -9,6 +9,10 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    aws: disabled
+    docker: enabled
+    none: enabled
```
s12chung: think it's safe to delete this addon.

templates/localstorageclass.yaml
joejulian: Perhaps so, but that would fall outside of the domain of this PR.

shaneutt (approved):

templates/awsstorageclass.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    aws: disabled
```
hectorj2f: aws is disabled ?

templates/dex-k8s-authenticator.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    all: disabled
```
hectorj2f: all disabled ? I don't get it

joejulian: I'll add more documentation in the kubeaddons library.

PR #76, opened 2019-07-05
Ops Portal authentication and authorization

jongiddy:
Configuration for the traefik-forward-auth authorizer, providing Ops Portal authentication and authorization.

### DIFF
--- a/templates/dex.yaml
+++ b/templates/dex.yaml
@@ -67,13 +67,14 @@ spec:
         oauth2:
           skipApprovalScreen: true
         staticClients:
-        # This `id` must by in sync with `dex-k8s-authenticator.yaml` value as well as
-        # kube-apiserver flag `oidc-client-id`.
+        # `redirectURIs` and `secret` values are modified in `configureDexStaticClients`
         - id: kube-apiserver
-          redirectURIs:
-            # This URI is just a placeholder and it will be replaced during the
-            # deployment with URL pointing to the traefik ingress public
-            # load balancer.
-            - 'https://dex-k8s-authenticator-kubeaddons.kubeaddons.svc.cluster.local:5555/token/callback/kubernetes-cluster'
+          # This `id` must by in sync with `dex-k8s-authenticator.yaml` value as well as
+          # kube-apiserver flag `oidc-client-id`.
           name: 'Kubernetes CLI authenticator'
-          # secret: value is generated automatically - see `setupDex()`
+          redirectURIs:
+            - 'https://PUBLIC.URI/token/callback/kubernetes-cluster'
+        - id: traefik-forward-auth
+          name: 'Ops Portal authenticator'
+          redirectURIs:
+            - 'https://PUBLIC.URI/_oauth'
--- a/templates/traefik-forward-auth.yaml
+++ b/templates/traefik-forward-auth.yaml
@@ -45,9 +45,9 @@ spec:
               key: username
         extraConfig: |
           rule.path.action = allow
-          rule.path.rule = Path(`/favicon.ico`)
+          rule.path.rule = Path(`/favicon.ico`, `/token`)
           rule.prefix.action = allow
-          rule.prefix.rule = PathPrefix(`/dex/`, `/dex-k8s-authenticator`)
+          rule.prefix.rule = PathPrefix(`/dex/`, `/token/`)
 
       ingress:
         enabled: true
--- a/templates/traefik-forward-auth.yaml
+++ b/templates/traefik-forward-auth.yaml
@@ -1,0 +1,61 @@
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: traefik-forward-auth
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: traefik-forward-auth
+    repo: https://mesosphere.github.io/charts/staging
+    version: 0.1.1
+    values: |
+      ---
+      replicaCount: 1
+
+      image:
+        repository: jongiddy/traefik-forward-auth
+        tag: latest
+        pullPolicy: Always
+
+      service:
+        type: ClusterIP
+        port: 4181
+
+      traefikForwardAuth:
+        oidcUri: "https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex"
+        clientId: traefik-forward-auth
+        clientSecret:
+          valueFrom:
+            secretKeyRef:
+              name: dex-client-secret-traefik-forward-auth
+              key: client_secret
+        caCertificate:
+          valueFrom:
+            secretKeyRef:
+              name: dex-root-ca
+              key: tls.crt
+        cookieSecure: true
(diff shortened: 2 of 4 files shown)


templates/traefik-forward-auth.yaml
```diff
@@ -0,0 +1,51 @@
...
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: https://github.com/jongiddy/traefik-forward-auth-chart/archive/oidc-issuer.tar.gz
+    version: 0.1.1
+    values: |
+      ---
+      replicaCount: 1
+
+      image:
+        repository: jongiddy/traefik-forward-auth
```
hectorj2f: when can we get it into a `mesosphere/traefik-forward-auth` repo or into `mesosphere/charts` 🙏  ?

templates/traefik-forward-auth.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: traefik-forward-auth
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  chartReference:
+    chart: https://github.com/jongiddy/traefik-forward-auth-chart/archive/oidc-issuer.tar.gz
```
hectorj2f: As below, could we get a PR with the chart in `mesosphere/charts` ?

templates/traefik.yaml
```diff
@@ -51,3 +51,8 @@ spec:
         defaultCN: traefik.localhost.localdomain
         defaultSANList:
           - "*"
+      forwardAuth:
+        entryPoints:
+        - https
+        address: http://traefik-forward-auth-kubeaddons.kubeaddons.svc.cluster.local:4181/
```
hectorj2f: Would traefik still work if we didn't enabled `dex` and all related dependencies ?

templates/traefik-forward-auth.yaml
jongiddy: This will take a little time. It's not just creating the GitHub repo. I also need to get it uploaded to Mesosphere's Docker repo.

templates/traefik-forward-auth.yaml
jongiddy: This should be faster. I want to remove the last vestiges of Konvoy-specific elements from the chart (the references to "well-known" secrets) and then I will add a PR to `mesosphere/charts`.

templates/traefik.yaml
jongiddy: No. This creates a dependency trail: traefik -> traefik-forward-auth -> dex. If any of these are unavailable, the Ops Portal is unusable.

What other paths do you think should be supported? Basic auth, no auth?

templates/traefik-forward-auth.yaml
jieyu: I think it's OK to for now while we're working on porting it to `mesosphere/traefik-forward-aut`

templates/traefik.yaml
jieyu: Have you tried to use `ingress.kubernetes.io/auth-url` label on the Ingress object? I think we should only forward auth on known paths for `opsportal` and `dex-k8s-authenticator`, instead of forcing the same auth for all?

jieyu: cc @joejulian here as well

jongiddy: I'll look into it. I was not aware that Traefik supported forward authentication on some paths only, because authentication occurs before any routing decisions are taken. I can't find any specifics in the Traefik docs, but https://stackoverflow.com/questions/50253357/add-basic-auth-for-admin-path-prefix suggests that it can be done (for basic auth at least). Note, currently, the rules provided in the `traefik-forward-auth` values spec define paths that are allowed without authentication.

templates/traefik-forward-auth.yaml
jongiddy: Since this chart is very specific to the Konvoy setup, with secrets obtained from other components, is it perhaps better to replace the Helm chart with a Kubernetes spec?

templates/traefik-forward-auth.yaml
hectorj2f: @jongiddy sounds good 👍, but please let's add a jira ticket and a comment to do it in the future.

templates/traefik-forward-auth.yaml
hectorj2f: We can go for now with this option 😞 to get this IN. But please, create a PR to mesosphere/charts so we can do the changes there.

jongiddy: https://github.com/mesosphere/charts/pull/16

templates/traefik-forward-auth.yaml
jongiddy: https://jira.mesosphere.com/browse/DCOS-56251

templates/traefik-forward-auth.yaml
hectorj2f: Thanks

templates/dex.yaml
```diff
@@ -65,6 +65,11 @@ spec:
...
             # load balancer.
-            - 'https://dex-k8s-authenticator-kubeaddons.kubeaddons.svc.cluster.local:5555/dex-k8s-authenticator/callback/kubernetes-cluster'
+            - 'https://PUBLIC.URI/dex-k8s-authenticator/callback/kubernetes-cluster'
           name: 'Kubernetes CLI authenticator'
           # secret: value is generated automatically - see `setupDex()`
+        - id: traefik-forward-auth
+          redirectURIs:
+            - 'https://PUBLIC.URI/_oauth'
+          name: 'Ops Portal authenticator'
+          # secret: ZXhhbXBsZS1hcHAtc2VjcmV0
```
hectorj2f: umm, I hope this isn't the value of the secret :). If we don't need it we could remove it.

jongiddy: All secrets get generated in `kubeaddons`. I will clean this up.

templates/traefik.yaml
jongiddy: I just want to note that the change to use ingress labels will mean we default to allowing anonymous access, with deny for specific pages. It is usually better to default to denying access and then carefully allow access, as the current `traefik-forward-auth` whitelist does. This makes it more difficult to add a new path and mistakenly leave it open for access.

hectorj2f: :/. okay let me to check it too :(

hectorj2f: As discussed yesterday, we might initially go with this approach for now.

jongiddy: I had a quick attempt at using `ingress.kubernetes.io/auth-url` but it did not work. So I'd like to merge as is and then look at the change separately, unless that is going to break anything now.

hectorj2f:
@shaneutt can you merge and generate a tag?

jongiddy:
I'm going to test this with a different ruleset that provides the authentication for `/ops/portal` and below, but no authentication elsewhere. This achieves the desired behavior, without needing to resolve the correct use of `ingress`

PR #80, opened 2019-07-09
kommander: be a chart

hectorj2f:
Move to use mesosphere/charts

### DIFF
--- a/templates/kommander.yaml
+++ b/templates/kommander.yaml
@@ -3,150 +3,17 @@ apiVersion: kubeaddons.mesosphere.io/v1alpha1
 kind: Addon
 metadata:
   name: kommander
-  namespace: kubeaddons
+  namespace: kommander
   labels:
     kubeaddons.mesosphere.io/name: kommander
 spec:
-  manifest: |
-    ---
-    apiVersion: v1
-    kind: Namespace
-    metadata:
-      name: kommander
-    spec:
-      finalizers:
-      - kubernetes
-    ---
-    apiVersion: apiextensions.k8s.io/v1beta1
-    kind: CustomResourceDefinition
-    metadata:
-      name: observableclusters.stable.mesosphere.com
-    spec:
-      group: stable.mesosphere.com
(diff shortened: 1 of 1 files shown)


hectorj2f:
@DanielMSchmidt question: why would I need to install multiple instances of kommander in different namespaces in my cluster 🤔 ? You could pass the desired namespace when installing it via Helm, but in kubeaddons-configs we are setting the namespace to be `kommander`.

DanielMSchmidt:
@hectorj2f The scenario we try to enable is to run one instance for Konvoy UI and one for the normal kommander usage. The other one would be if you need to split it up by organization

hectorj2f:
@DanielMSchmidt :/, that increases the complexity based on how we install addons using helm 🤔. I will create a ticket to see how we can address this request. This PR aims to move kommander code to a chart based on the existing scripts

DanielMSchmidt:
Sorry I think there was a misunderstanding.i would like to have the namespace configurable in the helm chart, but always install it with the same namespace in konvoy. I would like to be able to aim users towards an instantiation process for their second instance using helm and in this case it would be configurable. I understand this repo to be universal for helm installations, not bound to konvoy

joejulian:
tiller will create a namespace when you helm install to a namespace, ie:
"helm install --namespace fubar stable/minecraft" will create the
namespace "fubar" if it does not exist. We should not put namespace
records in helm charts, nor should we add namespace to the metadata. We should use the helm's namespace management directly.

joejulian:
"an instantiation process for their second instance" would look like:
```
helm install --namespace my_kommander mesospher_stable/kommander
```

DanielMSchmidt:
Ah, so the namespace option in helm sets every namespace in the kubernetes objects to be that namespace? That sounds awesome, I wish they had the feature back when I was working with it :D (or they had and we didn't know)

hectorj2f:
@DanielMSchmidt that depends on the year you were working with it :)

PR #89, opened 2019-07-10
chore: reduce default resource sizing

jieyu:
So that the default setting would allow us to fit all addons on a single
laptop.

Tested on macOS. we can do a default install with docker provisioner
with all addon enabled (assuming we change number of worker node
to 2 by default, which we should do)

### DIFF
--- a/templates/dashboard.yaml
+++ b/templates/dashboard.yaml
@@ -28,5 +28,5 @@ spec:
           cpu: 1000m
           memory: 1000Mi
         requests:
-          cpu: 500m
-          memory: 300Mi
+          cpu: 50m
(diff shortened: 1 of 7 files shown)


hectorj2f:
As far as I remember the general policy we used for DC/OS packages was to define some default settings for production environment. This default settings might become really problematic.

hectorj2f:
I'll rather prefer to have a dev directory in konvoy with a `cluster.yaml` configuration which fits on a docker cluster.

jieyu:
Per discussion, we decided to wait and do provider specific defaults once that feature is there.

PR #93, opened 2019-07-10
fix annotation typo

s12chung:


templates/opsportal.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: opsportal
   annotations:
-    kubeaddons.mesosphere.io/appVersion/opsportal: "1.0.0"
+    kubeaddons.mesosphere.io/opsportal: "1.0.0"
```
joejulian: Should be `appversion.kubeaddons.mesosphere.io/opsportal`

hectorj2f: 🤦‍♂

PR #97, opened 2019-07-11
Convert Velero To Helm Chart Implementation

shaneutt:
This PR converts Velero to a chart based implementation and utilizes [extrasteps](https://github.com/kubeaddons/extrasteps) to deploy a non-production fallback Minio cluster for the Velero Addon if no other backend is provided.

### DIFF
--- a/templates/velero.yaml
+++ b/templates/velero.yaml
@@ -1,3 +1,59 @@
+# ------------------------------------------------------------------------------
+# Velero
+#
+#
+# Velero is an open source backup and migration tool for Kubernetes.
+# See more about Velero at:
+#
+# * https://velero.io/
+# * https://github.com/heptio/velero
+# * https://github.com/helm/charts/tree/master/stable/velero
+#
+#
+# Implementation
+#
+#
+# Our implementation of Velero currently supports S3 backends for storage, and by default if no configuration overrides are
+# provided to point it at a backend other than the default, we will create and manage a distributed Minio (https://min.io/)
+# cluster which uses the default storage class for the cluster to maintain the backups.
+#
+# WARNING: using the default (fallback) backend is for testing purposes only and should not be used in production.
+#
+#
+# Testing
+#
+#
+# If you would like to test this Addon without deploying the suite of Addons you can deploy it to any cluster with the following:
+#
+#   export VELERO_VALUES="$(perl -MYAML::XS -e 'local $/; $y = Load <>; print $y->{spec}{chartReference}{values}' templates/velero.yaml)"
+#   export VELERO_VERSION="$(perl -MYAML::XS -e 'local $/; $y = Load <>; print $y->{spec}{chartReference}{version}' templates/velero.yaml)"
+#   echo "$VELERO_VALUES" | helm install --name velero --namespace velero stable/velero --version $VELERO_VERSION -f -
+#
+# Watch the pods and wait for everything to fully come online with:
+#
+#   kubectl -n velero get pods -w
+#
+# To use Velero via its' CLI tool, you'll need to download the tool from the repo:
+#
+#  * https://github.com/heptio/velero/releases
+#
+# You can view the default schedules with:
+#
+#   velero get schedules
+#
+# And schedule an initial backup with:
+#
+#   velero backup create initial-backup
+#
+# NOTE: By default we do not expose Velero or any of its components (e.g. Minio) outside of the cluster for security reasons.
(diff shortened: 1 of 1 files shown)


templates/velero.yaml
```diff
@@ -7,391 +63,40 @@ metadata:
...
+          config:
+            region: "fallback"
+      credentials:
+        secretContents:
+          cloud: "placeholder"
+      schedules:
+        default:
+          schedule: "0 0 * * *"
+      initContainers:
+        - name: initialize-velero
```
shaneutt:
```suggestion
        # initialize-velero manages deploying a fallback backend (e.g. Minio) for Velero if needed,
        # and also will manage the credentials and secrets for that backend and Velero automatically.
        - name: initialize-velero
```

shaneutt:
squashed

templates/velero.yaml
```diff
@@ -7,391 +63,42 @@ metadata:
...
-            - "mc --config-dir=/config config host add velero http://minio:9000 minio CHANGEME && mc --config-dir=/config mb -p velero/velero"
-            volumeMounts:
-            - name: config
-              mountPath: "/config"
+  chartReference:
+    chart: stable/velero
+    version: 2.1.1
+    values: |
+      ---
+      configuration:
```
gracedo: Can we enable metrics exporting? There is a `metrics.enabled` config field that is defaulted to `false`

shaneutt: Sure, sounds good :+1:

shaneutt: feb57b8bc82d7b8007d448069af3e897cb3c10f6 takes care of enabling metrics, and enabling the serviceMonitor for the Prometheus operator. Let me know if there's anything else you'd like tweaked before we merge @gracedo :+1:

templates/velero.yaml
```diff
@@ -1,3 +1,59 @@
+# ------------------------------------------------------------------------------
```
hectorj2f: Please, let's simply add a link to the Velero helm, this is making this YAML hard to read and maintain :/.

hectorj2f (changes requested):
Let's remove all that commented code in the yaml please 🙏 

Question: What is the size of the minio volume ?

templates/velero.yaml
shaneutt: > Let's remove all that commented code in the yaml please

Why?

> Question: What is the size of the minio volume?

[10g](https://github.com/kubeaddons/extrasteps/blob/master/pkg/minio/minio.go#L164) per instance (default 4 instances)

> Please, let's simply add a link to the Velero helm, this is making this YAML hard to read and maintain :/.

I'm confused by this assertion. Can you help me to better understand your take on this please?

gracedo (dismissed):

templates/velero.yaml
s12chung: i'd prefer a link to, but it's a non-blocker.

s12chung (dismissed):

templates/velero.yaml
hectorj2f: What is the value of adding all these text ? In my opinion, the value could be provided by a link to the velero helm and minio operator. There you can add the README and additional explanations about this addon. Let's keep this file as short as possible.

YOu could add this text to the mesosphere/chart of velero.

hectorj2f: Otherwise we could add a docs section here for velero. However I think that a link would be helpful enough.

shaneutt: I must respectfully disagree with your position on this, however since this is a blocker and holding up other work I've made a compromise and removed the testing section which included hints about how to deploy the addon:

a25e1eaa83c79961b87dacc5d6b1eaaa48b0e82b

I've also created a JIRA which covers the need to testing tooling for addons such as this was attempting to provide a little bit of for velero:

[DCOS-56456](https://jira.mesosphere.com/browse/DCOS-56456)

Let me know if this is an acceptable compromise from your end. I will mark this as resolved, please mark it as unresolved again if there is more you feel needs to be done.

gracedo (approved):

templates/velero.yaml
s12chung: I'm good if hector is ok. I still prefer a link.

shaneutt: What link @s12chung? There are three links below did you want something else added or did you mean JUST a link?

hectorj2f (approved):

PR #103, opened 2019-07-13
grafana: use X-Forwarded-User for the auth header

jieyu:
This is the header that traefik auth forwarder is using.
https://github.com/thomseddon/traefik-forward-auth#forwarded-headers


hectorj2f:
I tested and it works

PR #110, opened 2019-07-15
Convert Velero to Chart

shaneutt:
Converts Velero to Helm Chart implementation.

### DIFF
--- a/templates/velero.yaml
+++ b/templates/velero.yaml
@@ -1,3 +1,24 @@
+# ------------------------------------------------------------------------------
+# Velero
+#
+#
+# Velero is an open source backup and migration tool for Kubernetes.
+# See more about Velero at:
+#
+# * https://velero.io/
+# * https://github.com/heptio/velero
+# * https://github.com/helm/charts/tree/master/stable/velero
+#
+#
+# Implementation
+#
+#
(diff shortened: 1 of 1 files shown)


templates/velero.yaml
```diff
@@ -7,391 +28,47 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: velero
 spec:
+  namespace: velero
   kubernetes:
     minSupportedVersion: v1.14.0
   requires:
     matchLabels:
       kubeaddons.mesosphere.io/provides: defaultstorageclass
```
joejulian: Since the values are all aws specific, let's add this addon to the configs of the other cloud providers, but leave it disabled:

```suggestion
      kubeaddons.mesosphere.io/provides: defaultstorageclass
    cloudProvider:
    - name: aws
      enabled: true
    - name: docker
      enabled: false
    - name: none
      enabled: false
```

joejulian (changes requested):

templates/velero.yaml
s12chung: we add default values to other charts in kubeaddons (like metallb). is it possible to do that for this chart?

joejulian: and I don't know if that suggestion thing is working. I show six spaces before `kubeaddons` before the suggestion and after.

joejulian (approved):

PR #114, opened 2019-07-16
revert opsportal to manifest

alejandroEsc:


alejandroEsc:
> LGTM. Would appreciate some description in the PR for context next time.

Good point, will do so next time. Needed to unblock @shaneutt and address the fact that kubeaddons is not ready for this change yet.

PR #115, opened 2019-07-16
fixes cloudProvider values found in specs

alejandroEsc:
fixed velero so that is
- enabled in aws
- enabled in none
- not and addon in docker type.

fixed localpath and volume provisioners
- removed from aws deployment

prometheusadapter
- should reflect options found in prometheus

### DIFF
--- a/templates/localpathprovisioner.yaml
+++ b/templates/localpathprovisioner.yaml
@@ -14,7 +14,7 @@ spec:
     minSupportedVersion: v1.14.0
   cloudProvider:
     - name: none
-      enabled: true
+      enabled: false
   chartReference:
     chart: localpathprovisioner
     repo: https://mesosphere.github.io/charts/stable
--- a/templates/localpathprovisioner.yaml
+++ b/templates/localpathprovisioner.yaml
@@ -12,6 +12,9 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    - name: none
+      enabled: true
   chartReference:
     chart: localpathprovisioner
     repo: https://mesosphere.github.io/charts/stable
--- a/templates/localvolumeprovisioner.yaml
+++ b/templates/localvolumeprovisioner.yaml
@@ -13,8 +13,6 @@ spec:
   kubernetes:
     minSupportedVersion: v1.14.0
   cloudProvider:
-    - name: aws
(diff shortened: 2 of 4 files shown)


templates/localpathprovisioner.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.14.0
+  cloudProvider:
+    - name: none
+      enabled: true
```
hectorj2f: Note: The default should be localvolumeprovisioner for none, but I believe we want to let people choose the localpath as an option (although not enabled by default).

hectorj2f: These is always confusing me, so if this property means that it'll appear enabled then we don't want it because it will create a conflict with the localvolumeprovisioner.

alejandroEsc: I dont mind making that change, as long as we have concensus?

templates/localvolumeprovisioner.yaml
```diff
@@ -13,8 +13,6 @@ spec:
   kubernetes:
     minSupportedVersion: v1.14.0
   cloudProvider:
-    - name: aws
```
hectorj2f: We could use the localvolumeprovisioner on AWS but it should not be the enabled option by default

alejandroEsc: the problem here is that it would default the storage class to true? right?

templates/localpathprovisioner.yaml
hectorj2f: I shared what it was discussed in the storage meetings 🤔 , localvolumeprovisioner should be the default for none unless the user specifies the opposite.

templates/localvolumeprovisioner.yaml
hectorj2f: yes, i might be confused with this property (meaning enabled == it can be used but not enabled as the default one).

templates/localpathprovisioner.yaml
hectorj2f: Conclusion: Please, set the value to false

templates/localvolumeprovisioner.yaml
hectorj2f: Conclusion: let's keep it false and let the user to enable it themselves

PR #117, opened 2019-07-16
Convert OpsPortal to Chart Implementation

shaneutt:


shaneutt:
We no longer need this, as we're going to deprecate the older opsportal implementation

PR #124, opened 2019-07-17
Minor changes on Traefik's manifest

dduportal:
Hi there, 

while reviewing your product, I stumbled across 2 minors thing in the Traefik's default manifest:

* The link to the documentation should point explicitly to the current stable `v1.7`, to avoid breaking when the brand new Traefik v2.0 will be released.
* There was a duplicated YAML stanza about prometheus metrics enablement.

Thanks for your work!

### DIFF
--- a/templates/traefik.yaml
+++ b/templates/traefik.yaml
@@ -26,9 +26,6 @@ spec:
       service:
         labels:
           servicemonitor.kubeaddons.mesosphere.io/path: "metrics"
-      metrics:
-        prometheus:
-          enabled: true
       resources:
         limits:
           cpu: 2000m
--- a/templates/traefik.yaml
+++ b/templates/traefik.yaml
@@ -9,7 +9,7 @@ metadata:
   annotations:
     appversion.kubeaddons.mesosphere.io/traefik: "1.68.4"
     endpoint.kubeaddons.mesosphere.io/traefik: "/ops/portal/traefik"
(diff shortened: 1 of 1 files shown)


shaneutt:
Thanks @dduportal!

templates/traefik.yaml
```diff
@@ -26,9 +26,6 @@ spec:
       service:
         labels:
           servicemonitor.kubeaddons.mesosphere.io/path: "metrics"
-      metrics:
```
jieyu: Hum, can you explain more about this change?

Does that mean `/metrics` won't be served by Traefik?

hectorj2f: Yeah, why do you remove this change?

hectorj2f: This will disable traefik of exposing metric's

dduportal: The whole stanza
```yml
```
was duplicated, you can see the 2nd one here: https://github.com/mesosphere/kubeaddons-configs/pull/124/files#diff-76e7e224155bfbf77cdffd73534adae5L39 

(Yeah, Github diff view is not helping here, as you have to unfold the code view somewhere there is not a diff).

hectorj2f: oh! I see @dduportal, yeah GH didn't help here. 👍

dduportal: @hectorj2f no problem, thanks for the eagle eyes here!

PR #125, opened 2019-07-18
traefik-forward-auth: allow access to landing page

jieyu:
No auth needed for the landing page.

### DIFF
--- a/templates/kommander.yaml
+++ b/templates/kommander.yaml
@@ -23,16 +23,6 @@ spec:
     chart: kommander
     repo: https://mesosphere.github.io/charts/stable
     version: 0.1.7
-    values: |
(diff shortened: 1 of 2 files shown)


shaneutt:
@jieyu need signed commits and then :heavy_check_mark:

alejandroEsc:
@jieyu pr needs to be signed.

jongiddy:
This almost works, but makes the path `/ops/portal` unauthenticated. The second component of the rule name needs to be a unique identifier.

I can create a PR to fix this, but does the landing page need to be in the `/ops/portal` path? Since it selects whether you are going to the Ops Portal or getting a token, it conceptually seems to live outside that path.

PR #128, opened 2019-07-19
add konvoy-ui and opsportal config

s12chung:
depends on https://github.com/mesosphere/charts/pull/52.

also should fix annotations, once we get the docker image tag.

### DIFF
--- a/templates/konvoy-ui.yaml
+++ b/templates/konvoy-ui.yaml
@@ -24,8 +24,6 @@ spec:
     repo: https://mesosphere.github.io/charts/stable
     version: 0.1.10
     values: |
-      image:
-        tag: 1.47.2
       mode: konvoy
       logoutRedirectPath: /landing
       service:
--- a/templates/konvoy-ui.yaml
+++ b/templates/konvoy-ui.yaml
@@ -2,15 +2,14 @@
 apiVersion: kubeaddons.mesosphere.io/v1alpha1
 kind: Addon
 metadata:
-  name: kommander
+  name: konvoy-ui
   namespace: kubeaddons
   labels:
-    kubeaddons.mesosphere.io/name: kommander
+    kubeaddons.mesosphere.io/name: konvoy-ui
   annotations:
-    appversion.kubeaddons.mesosphere.io/kommander: "1.47.2"
-    endpoint.kubeaddons.mesosphere.io/kommander: /ops/portal/kommander
+    appversion.kubeaddons.mesosphere.io/konvoy-ui: "1.47.2"
+    endpoint.kubeaddons.mesosphere.io/konvoy-ui: /ops/portal/
 spec:
-  namespace: kommander
   kubernetes:
     minSupportedVersion: v1.15.0
   cloudProvider:
@@ -24,6 +23,15 @@ spec:
     chart: kommander
     repo: https://mesosphere.github.io/charts/stable
     version: 0.1.10
+    values: |
+      image:
+        tag: 1.47.2
+      mode: konvoy
+      logoutRedirectPath: /landing
+      service:
+        name: konvoy-ui
+      ingress:
+        path: /ops/portal
   requires:
     matchLabels:
       kubeaddons.mesosphere.io/provides: defaultstorageclass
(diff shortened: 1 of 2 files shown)


templates/konvoy-ui.yaml
```diff
@@ -0,0 +1,37 @@
...
+    appversion.kubeaddons.mesosphere.io/konvoy-ui: "1.24.2"
+    endpoint.kubeaddons.mesosphere.io/konvoy-ui: /ops/portal/
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
+  cloudProvider:
+    - name: aws
+      enabled: true
+    - name: docker
+      enabled: true
```
hectorj2f: I believe we don't need it enabled for docker

templates/konvoy-ui.yaml
```diff
@@ -0,0 +1,234 @@
...
+  name: opsportal
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: opsportal
+  annotations:
+    appversion.kubeaddons.mesosphere.io/opsportal: "1.0.0"
+spec:
+  kubernetes:
+    minSupportedVersion: v1.15.0
+  manifest: |
```
shaneutt: Manifest based addons are no longer supported @s12chung, do we need to convert this to a helm chart?

shaneutt: Manifest based addons are no longer supported @s12chung, do we need to convert this to a helm chart?

s12chung: I didn't add that. see the first commit.

templates/konvoy-ui.yaml
```diff
@@ -0,0 +1,37 @@
...
+      enabled: true
+    - name: none
+      enabled: true
+  chartReference:
+    chart: kommander
+    repo: https://mesosphere.github.io/charts/stable
+    version: 0.1.10
+    values: |
+      image:
+        tag: 1.47.2
```
hectorj2f: @s12chung I added the latest docker image tag, instead of `relpath`

hectorj2f: 🤞

s12chung: it's the default value, we can remove this: https://github.com/mesosphere/charts/blob/dev/stable/kommander/values.yaml#L3

forgot to remove it when I pushed up. my bad.

templates/konvoy-ui.yaml
```diff
@@ -0,0 +1,37 @@
...
+kind: Addon
+metadata:
+  name: konvoy-ui
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: konvoy-ui
+  annotations:
+    appversion.kubeaddons.mesosphere.io/konvoy-ui: "1.47.2"
+    endpoint.kubeaddons.mesosphere.io/konvoy-ui: /ops/portal/
+spec:
```
shaneutt: Just a reminder that if no `namespace` is provided in the spec, it will default to `kubeaddons. I believe that's what was intended here however.

templates/konvoy-ui.yaml
```diff
@@ -0,0 +1,37 @@
...
+      enabled: true
+    - name: docker
+      enabled: true
+    - name: none
+      enabled: true
+  chartReference:
+    chart: kommander
+    repo: https://mesosphere.github.io/charts/stable
+    version: 0.1.10
+    values: |
```
shaneutt: @s12chung just double checking, previously with the `kommander` chart no Values needed to be set as the defaults were all good. It would appear that's no longer the case?

templates/opsportal.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: opsportal
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: opsportal
+  annotations:
+    appversion.kubeaddons.mesosphere.io/opsportal: "1.0.0"
+spec:
+  kubernetes:
+    minSupportedVersion: v1.14.0
```
shaneutt:
```suggestion
    minSupportedVersion: v1.15.0
```

templates/konvoy-ui.yaml
s12chung: was intended :D

templates/konvoy-ui.yaml
s12chung: this is for `konvoy-ui`, it's the same chart, but different settings. `kommander` kubeaddons-configs has no values still.

PR #130, opened 2019-07-19
dashboards: add a dashboard to display per-node volume space usage

nfnt:
This dashboards shows the volume data usage of persistent volume claims.
These claims can be filtered by storage class and instance.

volume_space_usage

JIRA: [DCOS-55516](https://jira.mesosphere.com/browse/DCOS-55516)

### DIFF
--- a/dashboards/localvolumeusage.yaml
+++ b/dashboards/localvolumeusage.yaml
@@ -1,0 +1,351 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: prometheus-operator-grafana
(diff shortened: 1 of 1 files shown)


jdef:
"Instance" refers to ... "node" ?

nfnt:
There's both, "instance" as well as "node". "instance" seems to be the internal IP, while "node" seems to be a hostname. We've used "instance" before in the volume provisioner dashboards, that's why I'm using it here again.

shaneutt:
@jdef do you have anything else you want to discuss on this before we merge?

PR #131, opened 2019-07-19
templates: fix min supported version

hectorj2f:
Fix the minimum supporter versions to point to v1.15.0


hectorj2f:
@alejandroEsc the minimum version should be aligned with our minimal supported version which will be for GA `v1.15`

alejandroEsc:
@hectorj2f ok, i thought we were supporting one version behind, not sure why.

PR #132, opened 2019-07-19
konvoy-ui: don't attempt to create the CRD

hectorj2f:
related to: https://github.com/mesosphere/charts/pull/53

### DIFF
--- a/templates/konvoy-ui.yaml
+++ b/templates/konvoy-ui.yaml
@@ -22,8 +22,9 @@ spec:
   chartReference:
     chart: kommander
     repo: https://mesosphere.github.io/charts/stable
-    version: 0.1.10
+    version: 0.1.11
     values: |
+      createObservableClusterCRD: false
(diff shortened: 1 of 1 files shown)


s12chung:
please move `konvoy-ui` to install after `kommander`, they're running in parallel: https://github.com/mesosphere/kubeaddons/blob/c6333b85d81a0eefe5a84444bd6fff93f99263e0/pkg/workflows/dependent_workflow.go#L36

s12chung:
otherwise, lgtm.

hectorj2f:
@s12chung what is the problem of installing them in parallel ? Please explain it to me

s12chung:
`op-portal` will try to create an instance of that CRD, so in a way, it's dependent on `kommander`. so they shouldn't be in parallel.

hectorj2f:
no, it is done here https://github.com/mesosphere/kubeaddons/pull/317

PR #141, opened 2019-07-22
do not create kommander crd

alejandroEsc:
- As discussed, we will create the CR from the kubeaddons extra steps, we should not and will not do so here for now.


s12chung:
let's merge when the extrastep is merged.

s12chung:
extra step requires this to pass CI, as CI requires this to be in master

PR #147, opened 2019-07-23
Export kommander metrics to Prometheus

gracedo:
Bump Kommander to the latest version which exports metrics, and add the kommander namespace to the Prometheus service monitor match configs to get the metric endpoints scraped.

image


gracedo:
The grafana dashboard work will be separate - we'll need to tie the installation of the dashboard to whether kommander is running on the cluster, so the dashboard will probably not be added to this repo (tbd).

PR #151, opened 2019-07-23
Add polling intervals for konvoy ui

natmegs:
Add polling intervals used in kommander chart as env vars. Companion PR to mesosphere/charts#69

### DIFF
--- a/templates/konvoy-ui.yaml
+++ b/templates/konvoy-ui.yaml
@@ -22,7 +22,7 @@ spec:
   chartReference:
     chart: kommander
     repo: https://mesosphere.github.io/charts/stable
-    version: 0.1.15
+    version: 0.1.16
     values: |
       createObservableClusterCRD: false
       mode: konvoy


templates/konvoy-ui.yaml
```diff
@@ -27,6 +27,8 @@ spec:
       createObservableClusterCRD: false
       mode: konvoy
       logoutRedirectPath: /ops/landing
+      clusterPollingInterval: 3000
```
joejulian: Since this is the default 
```
clusterPollingInterval: 3000
clientPollingInterval: 3000
```
in the [chart](https://github.com/mesosphere/charts/blob/bb2cbb5618810e5cb8b21303c072c69c7799183c/stable/kommander/values.yaml#L7-L8) this change is not needed.

joejulian:
You _can_ merge this if you feel strongly, but I would err on the side of simplicity and just close this.

templates/konvoy-ui.yaml
hectorj2f: +1 on @joejulian comment

natmegs:
Ah I didn't realize it would use the other values as defaults - I've removed them here and bumped the chart version. Thanks for the help @joejulian @hectorj2f!

PR #157, opened 2019-07-24
Update DKA and add initContainer setup

shaneutt:
Blocked on https://github.com/mesosphere/charts/pull/75
Blocked on https://github.com/mesosphere/kubeaddons-extrasteps/pull/4

### DIFF
--- a/templates/dex-k8s-authenticator.yaml
+++ b/templates/dex-k8s-authenticator.yaml
@@ -23,7 +23,7 @@ spec:
   chartReference:
     chart: dex-k8s-authenticator
     repo: https://mesosphere.github.io/charts/staging
-    version: 1.1.5
+    version: 1.1.6
     values: |
       ---
       ingress:
@@ -34,9 +34,6 @@ spec:
         hosts:
           - ""
       dexK8sAuthenticator:
-        port: 5555
-        debug: false
-        web_path_prefix: /token
         #logoUrl: http://<path-to-your-logo.png>
         #tlsCert: /path/to/dex-client.crt
         #tlsKey: /path/to/dex-client.key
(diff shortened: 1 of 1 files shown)


templates/dex-k8s-authenticator.yaml
```diff
@@ -34,22 +34,26 @@ spec:
...
-          # load balancer.
+          # This URI is just a placeholder and it will be replaced during initContainers
+          # with a URL pointing to the traefik ingress public load balancer.
           redirect_uri: https://dex-k8s-authenticator-kubeaddons.kubeaddons.svc.cluster.local:5555/token/callback/kubernetes-cluster
+      caCerts:
+        enabled: true
+        secrets:
+        - name: dex-root-ca
+          filename: dex-root-ca.crt
+          value: cGxhY2Vob2xkZXIK # b64 "placeholder" this will be updated by initcontainers
```
hectorj2f: `value: cGxhY2Vob2xkZXIK ` I think this is wrong...

shaneutt:
```
$ echo 'cGxhY2Vob2xkZXIK' |base64 -d
placeholder
```

joejulian (approved):

templates/dex-k8s-authenticator.yaml
hectorj2f: Why do we need in first place to add this secret ?

hectorj2f: That should be done by the init container, right ?

hectorj2f: that is what I assume instead of having to add a secret with values that will be overwritten afterwards.

shaneutt: This is related to https://github.com/kubernetes/kubernetes/issues/35183, upstream Kubernetes does not currently have a solid mechanism for the opposite of an initContainer, and so the full lifecycle of the secret resource can not be managed by the underlying deployment. Adding it here makes helm responsible for the end-of-life of this resource and so it will be cleaned up properly as per helm delete.

hectorj2f: This is ugly 🤦‍♂. I am not envisioning a scenario where someone installs dex and then uninstalls it to keep a cluster working 🤔 .  I am still thinking this should be done by the init-container, but I won't block you.

shaneutt: In this particular case, this is explicitly done via values to workaround a deficiency in the helm chart. [DCOS-56979](https://jira.mesosphere.com/browse/DCOS-56979) is a follow up to deal with this deficiency as part of later iterations. This should help with the ugliness @hectorj2f.

shaneutt: Talked to hector in Slack, going to consider this resolved.

PR #159, opened 2019-07-25
[chore] allow the creation of the crd

alejandroEsc:
need to allow creation of crd.


joejulian:
yeah, that might help

PR #162, opened 2019-07-25
awsebscsiprovisioner: bump the version

hectorj2f:


shaneutt:
Just confirming the chart is published before we merge?

nfnt:
There is a chart at https://github.com/mesosphere/charts/blob/master/docs/stable/awsebscsiprovisioner-0.2.8.tgz and it is referenced in https://github.com/mesosphere/charts/blob/master/docs/stable/index.yaml

PR #168, opened 2019-07-26
Utilize stable for addon-initializer

shaneutt:
https://github.com/mesosphere/kubeaddons-configs/pull/168#issuecomment-515971507

### DIFF
--- a/templates/dex-k8s-authenticator.yaml
+++ b/templates/dex-k8s-authenticator.yaml
@@ -58,5 +58,5 @@ spec:
           value: cGxhY2Vob2xkZXIK # b64 "placeholder" this will be updated by initcontainers
       initContainers:
       - name: initialize-dka-config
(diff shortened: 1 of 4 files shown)


s12chung (changes requested):
we have to lock versions because when we release, we want to ensure the code is stable.

shaneutt:
Yep, sorry I should have added to the description that this is an interim solution to make sure all usage of the tool is testing the same version until GA, at which time we'll be moving the builds over to the `mesosphere/` org.

The point here is that over the next few days, all testing should be done on the same version of the extrasteps code and from that testing we will harden a release version (the tool has no actual release versions thus far), that way we're not testing different usages on different builds only to find out at the last minute that there are problems for some components in the latest stable build.

The follow up for this is https://jira.mesosphere.com/browse/DCOS-57126

shaneutt:
Actually going to simply close this for now, since we don't really have the time to soak this that I was hoping for.

PR #175, opened 2019-07-30
localvolumeprovisioner: change the namespace to kube-system

hectorj2f:
This is a proposal to overcome this issue https://github.com/kubernetes/kubernetes/issues/76308 when using priorityClassName: https://github.com/mesosphere/charts/blob/dev/stable/localvolumeprovisioner/values.yaml#L12

My intention is to treat the localvolumeprovisioner similarly to how it is treated another storage provisioner

### DIFF
--- a/templates/localvolumeprovisioner.yaml
+++ b/templates/localvolumeprovisioner.yaml
@@ -3,7 +3,7 @@ apiVersion: kubeaddons.mesosphere.io/v1alpha1
 kind: Addon
 metadata:
   name: localvolumeprovisioner
-  namespace: kube-system
+  namespace: kubeaddons
   labels:
     kubeaddons.mesosphere.io/name: localvolumeprovisioner
     kubeaddons.mesosphere.io/provides: storageclass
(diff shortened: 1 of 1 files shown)


templates/localvolumeprovisioner.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: kubeaddons.mesosphere.io/v1alpha1
 kind: Addon
 metadata:
   name: localvolumeprovisioner
-  namespace: kubeaddons
+  namespace: kube-system
```
dkoshkin: Thats the wrong namespace, see https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/awsebscsiprovisioner.yaml#L15

shaneutt: Yes, currently the addon itself will be in kubeaddons even if the actual deployment ends up in kube-system. Improvements for this are suggested in [https://jira.mesosphere.com/browse/DCOS-57136](DCOS-57136) which could potentially be the follow up to this for later iterations.

hectorj2f: good catch!

PR #185, opened 2019-07-31
Set resources for multiple addons

dkoshkin:
Blocked by: https://github.com/mesosphere/charts/pull/97

### DIFF
--- a/templates/metallb.yaml
+++ b/templates/metallb.yaml
@@ -21,4 +21,11 @@ spec:
     version: 0.9.5
     values: |
       ---
-      # we currently provide no additional configuration for metallb
+      controller:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      speaker:
+        requests:
+          cpu: 100m
+          memory: 100Mi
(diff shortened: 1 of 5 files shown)


templates/metallb.yaml
```diff
@@ -21,4 +21,11 @@ spec:
     version: 0.9.5
     values: |
       ---
-      # we currently provide no additional configuration for metallb
+      controller:
+        limits:
+          cpu: 100m
+          memory: 100Mi
+      speaker:
+        limits:
```
jieyu: set requests instead?

templates/metallb.yaml
```diff
@@ -21,4 +21,11 @@ spec:
     version: 0.9.5
     values: |
       ---
-      # we currently provide no additional configuration for metallb
+      controller:
+        limits:
```
jieyu: set requests instead?

templates/metallb.yaml
hectorj2f: good call

templates/metallb.yaml
hectorj2f: yep

hectorj2f:
I am going to merge this to start testing it

hectorj2f:
okay... i am signing the commits

PR #190, opened 2019-08-05
Enable velero service monitor

gracedo:
I read through some issues in the prometheus-operator helm chart and found [this thread](https://github.com/helm/charts/issues/13196) which explained how to get service monitors in other namespaces pulled in. I believe this effectively fixes https://jira.mesosphere.com/browse/DCOS-56470 as well.

image

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -67,6 +67,8 @@ spec:
                 port: metrics
                 interval: 30s
         prometheusSpec:
+          serviceMonitorSelectorNilUsesHelmValues: false
+          serviceMonitorNamespaceSelector: {}
           additionalScrapeConfigs:
             - job_name: 'kubernetes-nodes-containerd'
               metrics_path: /v1/metrics
--- a/templates/velero.yaml
+++ b/templates/velero.yaml
@@ -67,6 +67,6 @@ spec:
       metrics:
         enabled: true
         # TODO: re-enable the serviceMonitor later, it doesn't work out of the box with the current stable velero chart (see DCOS-56470)
-        # serviceMonitor:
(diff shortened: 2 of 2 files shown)


gracedo:
@shaneutt I think this makes sense, although I haven't yet run into this issue with velero. And I'm not super familiar with helm, but wouldn't this be an issue with any chart that installs service monitors? I'm wondering why we don't have this same issue with the other addons?

(I've been reading through https://github.com/helm/helm/issues/2994)

shaneutt:
This could in theory be an issue with any chart that installs a `ServiceMonitor` resource.

The problem domain for us is that we provide a validated configuration and audited selection of addons with default configurations (like enabling `ServiceMonitoring`, shortly). Regardless of how we implement it we're going to need to cover the situation where Velero is enabled for deployment but Prometheus is not.

Let us know if you find anything compelling upstream to indicate how this is handled (or, I suppose _not handled_) by other tools. We're at a point where there are several possible directions to take and so more information should serve to help us narrow that down, let us know what your preferred course of action is here and let's talk through the best solution.

gracedo:
https://github.com/mesosphere/charts/pull/104 The issue of CRD management is still outstanding, but separately I think the better way for us to get velero metrics discovered is to use service label matching on already-existing service monitors defined in prometheus instead of defining and using a new service monitor here.

We should follow and keep track of https://github.com/helm/helm/issues/5871 re: the CRD dependency and management issue

shaneutt:
Thanks @gracedo, I've created [DCOS-57432](https://jira.mesosphere.com/browse/DCOS-57432) to track this issue for follow up internally.

PR #207, opened 2019-09-04
Change how addons require a default StorageClass

shaneutt:
This PR removes the search for labels to find a provided default storage class, and instead simply flags whether it needs a default storage class as part of the requirements.

This will help kubeaddons to better support a wider range of clusters, and provide end users more options to determine where the storage for their addons comes from.

### DIFF
--- a/templates/elasticsearch.yaml
+++ b/templates/elasticsearch.yaml
@@ -6,6 +6,9 @@ metadata:
   namespace: kubeaddons
   labels:
     kubeaddons.mesosphere.io/name: elasticsearch
+    # TODO: we're temporarily supporting dependency on an existing default storage class
+    # on the cluster, this hack will trigger re-queue on Addons until one exists.
+    kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: true
   annotations:
     appversion.kubeaddons.mesosphere.io/elasticsearch: "6.7.0"
 spec:
@@ -18,9 +21,6 @@ spec:
       enabled: false
     - name: none
       enabled: true
-  requires:
-    matchLabels:
-      kubeaddons.mesosphere.io/provides: defaultstorageclass
   chartReference:
     chart: stable/elasticsearch
     version: 1.26.2
--- a/templates/kommander.yaml
+++ b/templates/kommander.yaml
@@ -6,6 +6,9 @@ metadata:
   namespace: kubeaddons
   labels:
     kubeaddons.mesosphere.io/name: kommander
+    # TODO: we're temporarily supporting dependency on an existing default storage class
+    # on the cluster, this hack will trigger re-queue on Addons until one exists.
+    kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: true
   annotations:
     appversion.kubeaddons.mesosphere.io/kommander: "1.50.0"
     endpoint.kubeaddons.mesosphere.io/kommander: /ops/portal/kommander
@@ -24,6 +27,3 @@ spec:
     chart: kommander
     repo: https://mesosphere.github.io/charts/stable
     version: 0.1.21
-  requires:
-    matchLabels:
-      kubeaddons.mesosphere.io/provides: defaultstorageclass
--- a/templates/traefik.yaml
+++ b/templates/traefik.yaml
@@ -6,6 +6,9 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: traefik
     kubeaddons.mesosphere.io/provides: ingresscontroller
+    # TODO: we're temporarily supporting dependency on an existing default storage class
+    # on the cluster, this hack will trigger re-queue on Addons until one exists.
+    kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: true
   annotations:
     appversion.kubeaddons.mesosphere.io/traefik: "1.68.4"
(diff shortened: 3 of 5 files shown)


joejulian (changes requested):
I'm against this

shaneutt:
@joejulian could you please articulate the "why" so we can have discourse on the subject.

s12chung:
I'm ok with this one. this plus the validation Webhook will handle all the cases.

alejandroEsc:
@joejulian why the disagreement? Would like to know more. I for one think this is ok and more explicit.

alejandroEsc (dismissed):

templates/kommander.yaml
```diff
@@ -25,5 +25,4 @@ spec:
     repo: https://mesosphere.github.io/charts/stable
     version: 0.1.21
   requires:
-    matchLabels:
-      kubeaddons.mesosphere.io/provides: defaultstorageclass
+    defaultStorageClass: true
```
jieyu: why Kommander requires default storage class?

jieyu (commented):
I think this is a reasonable approach.

I found two issues irrelevant to the PR>

templates/traefik.yaml
```diff
@@ -14,8 +14,7 @@ spec:
   kubernetes:
     minSupportedVersion: v1.15.0
   requires:
-    matchLabels:
-      kubeaddons.mesosphere.io/provides: defaultstorageclass
+    defaultStorageClass: true
```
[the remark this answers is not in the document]
hectorj2f: +1

templates/kommander.yaml
hectorj2f: +1

templates/elasticsearch.yaml
```diff
@@ -19,8 +19,7 @@ spec:
     - name: none
       enabled: true
   requires:
-    matchLabels:
-      kubeaddons.mesosphere.io/provides: defaultstorageclass
+    defaultStorageClass: true
```
joejulian: Since this is supposed to be a temporary workaround, we should not be adding fields to the struct that we'll have to take away. This will cause a breaking change to the API.

templates/traefik.yaml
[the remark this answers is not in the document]
joejulian: Because the chart has the ability to create a pvc.

templates/elasticsearch.yaml
shaneutt: We can accommodate for those changes with versioning and version to version transforms. I think this is probably one of our last opportunities to make such a change with low impact since we _still_ need to get a beta version of our API minted soon anyway.

templates/kommander.yaml
shaneutt: @DanielMSchmidt do you potentially have any comments on this? Right now we're just looking at potentially changing how this is defined, but while we're here is this actually a requirement for Kommander?

templates/elasticsearch.yaml
joejulian: I do not feel that kubeaddons should be getting into determining if specific primitives are installed or configured in a specific way. This opens the gate to allowing developers to check for any object at all. Do we want to let the controller watch for a specific Event before an addon can proceed? Do we want kubeaddons to check to see if the dex-secret exists and do something different with an Addon if it does? I don't believe we do want that.

I feel kubeaddons should only act on Addon resources. It should use a driver to resolve the state of those resources. That's it. It should not be smarter than that.

joejulian: It also begins to add a level of complexity to the controller that is unnecessary. This can be solved in several other ways.

templates/kommander.yaml
DanielMSchmidt: Not that I'm aware of, we don't really store anything on disk in kommander

templates/elasticsearch.yaml
shaneutt: > Do we want to let the controller watch for a specific Event before an addon can proceed? Do we want kubeaddons to check to see if the dex-secret exists and do something different with an Addon if it does? I don't believe we do want that.
> I feel kubeaddons should only act on Addon resources. It should use a driver to resolve the state of those resources. That's it. It should not be smarter than that.

I have a different take from this position. I'm for flexibility at the native Kubernetes API level. I am not opposed to enabling functionality that requires interaction with native, documented features of the Kubernetes API. We are already a Kubernetes native implementation, our API utilizes the provisions of Kubernetes and I see the default storageclass functionality as just another one of those provisions. I would be opposed to creating couplings between other APIs from other systems but not the native Kubernetes API.

templates/kommander.yaml
shaneutt: Thanks for the update Daniel.

I've created https://jira.mesosphere.com/browse/DCOS-58414 as a bug report to follow up on this, so as to ensure the change is tested in isolation and not added to the scope of this change.

templates/elasticsearch.yaml
shaneutt: We talked about this in standup and came to alignment, updates have been made regarding that alignment.

shaneutt:
@joejulian ready for your re-review with the changes we discussed in standup today: using a label-based hack for now to get this in place, but saving deeper implementation for later.

joejulian (dismissed):

joejulian (approved):

alejandroEsc (approved):

s12chung (approved):

PR #210, opened 2019-09-06
features: Kubefed

alejandroEsc:
- enable kubefed deployment 
  - enabled by default
  -  kommander requires this deployment
  - deployed in kommander namespace

### DIFF
--- a/templates/kommander.yaml
+++ b/templates/kommander.yaml
@@ -27,3 +27,6 @@ spec:
     chart: kommander
     repo: https://mesosphere.github.io/charts/stable
     version: 0.1.21
+  requires:
+    matchLabels:
+      kubeaddons.mesosphere.io/name: kubefed
--- a/templates/kubefed.yaml
+++ b/templates/kubefed.yaml
@@ -21,6 +21,6 @@ spec:
     - name: none
       enabled: true
   chartReference:
-    chart: kubefed-charts/kubefed
+    chart: kubefed
     repo: https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
     version: 0.1.0-rc6
--- a/templates/kubefed.yaml
+++ b/templates/kubefed.yaml
@@ -23,4 +23,4 @@ spec:
   chartReference:
     chart: kubefed-charts/kubefed
     repo: https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
-    version: v0.1.0-rc6
+    version: 0.1.0-rc6
--- a/templates/kubefed.yaml
+++ b/templates/kubefed.yaml
@@ -13,6 +13,13 @@ spec:
   kubernetes:
     minSupportedVersion: v1.15.0
   namespace: kommander
+  cloudProvider:
+    - name: aws
+      enabled: true
+    - name: docker
+      enabled: true
+    - name: none
+      enabled: true
   chartReference:
     chart: kubefed-charts/kubefed
     repo: https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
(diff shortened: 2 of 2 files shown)


jimmidyson:
Thanks for doing this @alejandroEsc! Can you share how to test it please because without automated tests I'd like to try it out myself before approving? Also, have you tested uninstallation (do we care about uninstallation? :thinking:), because kubefed uses finalizers a lot and uninstalling kubefed without deleting federated resource definitions first (or just removing finalizers) will cause issues like not being able to delete namespaces containing federated resources (finalizers will never be removed).

alejandroEsc:
@jimmidyson so to test was not simple since there are on going changes, however you can do the following (note this uses a branch from the v0.1.2 tag)

in the konvoy repo run:
```
$ go get github.com/mesosphere/kubeaddons@ae/kubefed2
```
then make the binary

```
$ make dist
```
 copy/move etc.. the binary and use it to run init

in the `cluster.yaml` file make the following additions:

```
...
  addons:
    configVersion: ae/kubefed
...
    - name: kubefed
      enabled: true
...
```

finally the result should be:


```
$ k get pods -n kommander
NAME                                          READY   STATUS    RESTARTS   AGE
kommander-kubeaddons-7b75fb7d6-qlzx8          1/1     Running   0          101m
kubefed-admission-webhook-7fdcc5b56f-xsb5n    1/1     Running   0          2m4s
kubefed-controller-manager-75bdd8c787-9gmb7   1/1     Running   0          2m4s
kubefed-controller-manager-75bdd8c787-rkz5h   1/1     Running   0          2m4s
```

Deletion: if you chose to disable the addon it does work. However, from this initial point no clusters have been 'joined' yet. Also this works fine with kommander since there are no hard requirements yet on being a dependency etc.. Finally, as you may have noticed, kubeaddons as is does require some changes to make this work.

jimmidyson:
@alejandroEsc OK I'm still a bit concerned about the uninstallation issue (obvisously after joining clusters and propagating stuff to those clusters) but perhaps it's enough to just document the uninstallation, and hope that people don't regularly uninstall without removing federated stuff first.

jimmidyson:
@alejandroEsc @shaneutt When considering dependencies, I don't see why there should be changes in `mesosphere/kubeaddons` directly? Reading https://github.com/mesosphere/kubeaddons#dependencies, it should be enough to mark `kubefed` as a dependency of `kommander`?

shaneutt:
@alejandroEsc  & @jimmidyson looking at Alejandro's branch it appears that is branched using the deprecated library version of Kubeaddons - we don't need to support this version to my knowledge?

Is the uninstallation issue even an issue with Kubeaddons master and the controller machinery?

jimmidyson:
> Is the uninstallation issue even an issue with Kubeaddons master and the controller machinery?

Yes (not because of kubeaddons) because this is how the chart leaves CRs (that rely on finalizers to be deleted) and CRDs behind when uninstalling. This is a common issue with helm charts and I hope will be resolved in helm3 perhaps?

shaneutt:
Ah, indeed. I'm not sure Helm V3 does anything useful in that regard, but I will investigate, we're close to the point where we should probably be adding a v3 driver for kubeaddons.

alejandroEsc:
> @alejandroEsc & @jimmidyson looking at Alejandro's branch it appears that is branched using the deprecated library version of Kubeaddons - we don't need to support this version to my knowledge?
> 
> Is the uninstallation issue even an issue with Kubeaddons master and the controller machinery?

@shaneutt @jimmidyson  I am trying to ensure I can test this kubeaddons-configs with existing konvoy which is using `github.com/mesosphere/kubeaddons v0.1.2` and not a controller. In this codebase the addon must be explicitly stated in a dep workflow list so in order to test i had to add it in. I will not be committing this UNLESS we somehow dont get the controller work out.

shaneutt:
cc @joejulian just so he's aware of this too

PR #212, opened 2019-09-10
nvidia: Added nvidia kubeaddon config

Gilbert88:
This addon includes nvidia drive container and device plugin.

Corresponding to https://github.com/mesosphere/charts/pull/120

### DIFF
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -15,7 +15,7 @@ spec:
   namespace: kube-system
   cloudProvider:
     - name: aws
-      enabled: true
+      enabled: false
     - name: docker
       enabled: false
     - name: none
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -24,3 +24,29 @@ spec:
     chart: nvidia
     repo: https://mesosphere.github.io/charts/staging
     version: 0.1.0
+    values: |
+      ---
+      nvidia-driver:
+        resources:
+          limits:
+             cpu: 2000m
+             memory: 6144Mi
+          requests:
+             cpu: 200m
+             memory: 256Mi
+        nodeSelector:
+          konvoy.mesosphere.com/gpu-provider: NVIDIA
+      nvidia-device-plugin:
+        resources:
+          limits:
+             cpu: 200m
+             memory: 128Mi
+          requests:
+             cpu: 100m
+             memory: 128Mi
+        nodeSelector:
+          konvoy.mesosphere.com/gpu-provider: NVIDIA
+        initContainers:
+        - name: init-wait
+          image: busybox
+          command: ['sh', '-c', 'sleep 90']
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -12,7 +12,7 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.15.0
-  namespace: kube-system 
+  namespace: kube-system
   cloudProvider:
     - name: aws
       enabled: true
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -1,0 +1,26 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: nvidia
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: nvidia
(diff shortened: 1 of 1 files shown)


shaneutt:
Looks cool @Gilbert88, if you need anything while you're working on this feel free to let us know and check in with us in the #kubeaddons channel on Slack! :smiley:

shaneutt:
@Gilbert88 :wave: similar to https://github.com/mesosphere/charts/pull/120 I just wanted to double check if there's any assistance we can be providing that would be helpful here, just let us know :clinking_glasses:

templates/nvidia.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: nvidia
```
hectorj2f: Does it depends on other addons to work?

Gilbert88: Nope, but it depends on the konvoy cluster config to enable it:

```
  nodePools:
  - name: worker
    gpu:
      nvidia:
        enabled: false
  - name: gilbert
    gpu:
      nvidia:
        enabled: true
```

If not turned on, it is no-op - the addon will not be installed on nodes that haven't enabled Nvidia GPU.

hectorj2f: We might need to validate that in kubeaddons

shaneutt: > We might need to validate that in kubeaddons

Kubeaddons proper treats the configuration of the Addons as opaque, so at least presently it does not have a feature for inline or automatic reconfiguration of Addons.

Gilbert88:
> As mentioned in [mesosphere/charts#120 (comment)](https://github.com/mesosphere/charts/pull/120#discussion_r330791010), please set some estimations in terms of resource usage. We can easily changed them here afterwards.

Spent some time to test out different combinations of default resources. The current estimate should be very close to what default resources will be in the future.

hectorj2f:
@Gilbert88 we are waiting to merge some important changes in kubeaddons (hopefully today) once those are added, we'll merge this PR to avoid adding more stuff that could potentially fail and add noise to the current addons code.

templates/nvidia.yaml
```diff
@@ -0,0 +1,52 @@
...
+    kubeaddons.mesosphere.io/provides: nvidia
+  annotations:
+    appversion.kubeaddons.mesosphere.io/nvidia: "0.1"
+spec:
+  kubernetes:
+    minSupportedVersion: v1.15.0
+  namespace: kube-system
+  cloudProvider:
+    - name: aws
+      enabled: true
```
hectorj2f: @shaneutt will this enable it by default when deploying on AWS ? yes, right!

shaneutt: From what perspective?

hectorj2f: Let's set `enabled: false` to be merge this PR asap. We won't enabled it until we get other changes in our master (kubeaddons controller)

jimmidyson: @hectorj2f i've just dismissed your review so this doesn't get accidentally merged until you're happy.

Gilbert88: @hectorj2f I can disable it be default.

However, FYI even if we enable the Nvidia Addon, daemonsets will not be installed unless GPU is enabled from cluster config:
```
  nodePools:
  - name: worker
    gpu:
      nvidia:
        enabled: false
  - name: gilbert
    gpu:
      nvidia:
        enabled: true
```

Because we have leveraged the nodeSelector and Konvoy has done the labeling part for GPU machines.

hectorj2f: We know that. I am asking to set enabled false here for aws just so we disabled it as an add-on to be attempted for deployment.

By the way the fact that Nvidia flag in cluster.yaml gets enabled, it doesn't mean this addon will be deployed. We need to handle that scenario in Konvoy or any component

Gilbert88: > By the way the fact that Nvidia flag in cluster.yaml gets enabled, it doesn't mean this addon will be deployed. We need to handle that scenario in Konvoy or any component

This is a good point. The addon depends on the cluster flag, and it does not mean the addon has to be deployed if the cluster flag is enabled. We can imagine a use case that a user expects that Konvoy handles the low-level container runtime (from the cluster flag) but they wanted to customize the nvidia driver and device plugin.

We don't necessarily need to enable the Nvidia addon by default!

PR #214, opened 2019-09-10
Add CONTRIBUTING.md

shaneutt:


CONTRIBUTING.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Customizing Existing Addons
+
+Addons can be customized via their manifests in the `templates/` directory.
+
+Changing an addons name or namespace is done in the `metadata` field.
+
+Configuration of the `Addon` resource itself is primarily done via via the `spec` field, particularly changing the [helm values configuration]() of a helm-based addon is done via the `spec.ChartReference.Values` field.
+
+## ExtraSteps
```
alejandroEsc: It is a bit more involved. Does this belong here? I suspect we may want to mention it but the actual details belong in  mesosphere/charts.

alejandroEsc (commented):
looks good overall

CONTRIBUTING.md
samba: Mentioning it here is important, but the details of _creating an init container_ should be in `kubeaddons-extrasteps`; perhaps mentioning them in `charts` is also a good idea, since they'll be referenced from there.

shaneutt: I'll be adding CONTRIBING.md files to these other two repos as well, I will ensure they cover their angles on this as well :+1:

shaneutt:
related https://github.com/mesosphere/charts/pull/121

hectorj2f (approved):

alejandroEsc (approved):

PR #216, opened 2019-09-19
Bump alertmanager version to v0.17.0

gracedo:
In preparation for running [Karma](https://github.com/helm/charts/tree/master/stable/karma) as a central alerting dashboard in Kommander, we need to bump Konvoy Alertmanagers up to a version > 0.16.x in order to be compatible with Karma (it is currently defaulted to v0.16.2 on our version of prometheus-operator). See https://github.com/prymitive/karma/issues/115 for more details

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -184,6 +184,8 @@ spec:
               memory: 1000Mi
       alertmanager:
         alertmanagerSpec:
+          image:
+            tag: v0.17.0
           resources:
             limits:
               cpu: 100m


gracedo:
@hectorj2f I was able to see that Alertmanager version bump didn't break anything, but didn't check any of the other components. I think we can wait to bump other components until there's a specific reason why we need to upgrade (case here was that Karma doesn't work with the version of Alertmanger we were on), or until we upgrade prometheus-operator chart which would contain component bumps that are tested and compatible with each other.

PR #218, opened 2019-09-23
awsebcsi: bump chart version

hectorj2f:


s12chung:
you need to bump: `version: 0.2.9` don't you?

templates/awsebscsiprovisioner.yaml
```diff
@@ -19,7 +19,7 @@ spec:
   chartReference:
     chart: awsebscsiprovisioner
     repo: https://mesosphere.github.io/charts/stable
-    version: 0.2.9
+    version: 0.4.0
```
s12chung: i believe this is incorrect: https://github.com/mesosphere/charts/pull/122/files#diff-89a57bca70adf41ec198f6e05b535c57L9

hectorj2f: thanks @s12chung good catch!

PR #219, opened 2019-09-24
Add helm values yaml metadata to some addons

shaneutt:
This PR adds some links to the annotations of several addons which indicate that the addon has default helm chart values and links to a view of those defaults.

Resolves [DCOS-59121](https://jira.mesosphere.com/browse/DCOS-59121)

### DIFF
--- a/templates/traefik-forward-auth.yaml
+++ b/templates/traefik-forward-auth.yaml
@@ -8,7 +8,8 @@ spec:
     minSupportedVersion: v1.15.0
   requires:
     matchLabels:
-      kubeaddons.mesosphere.io/name: dex    
+      kubeaddons.mesosphere.io/name: dex
+      values.chart.helm.kubeaddons.mesosphere.io/traefik-forward-auth: "https://raw.githubusercontent.com/mesosphere/charts/6c43b8ab10108fb1adba5c6dd10e800e5f1abdd0/staging/traefik-forward-auth/values.yaml"
   chartReference:
     chart: traefik-forward-auth
     repo: https://mesosphere.github.io/charts/staging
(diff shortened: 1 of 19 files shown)


s12chung:
think you want the raw file don't you?

joejulian:
Not only might you want the raw contents, but you should probably also get from the same version as the chart (basically an impossible task through git/github).

Probably the best way to get that would be by either processing the index.yaml, ie:
```
$ yq '.entries["awsebscsiprovisioner"][] | select(.version == "0.1.0").urls[0]' index.yaml
https://mesosphere.github.io/charts/stable/awsebscsiprovisioner-0.1.0.tgz
```
then pulling the values.yaml from the tarball, or using `helm fetch ` and pulling values.yaml from the tarball.

templates/traefik-forward-auth.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     minSupportedVersion: v1.15.0
   requires:
     matchLabels:
-      kubeaddons.mesosphere.io/name: dex    
+      kubeaddons.mesosphere.io/name: dex
```
hectorj2f: why didn't you add it here ?

shaneutt:
Rebasing and WIP for a bit

templates/traefik-forward-auth.yaml
shaneutt: will add this in the rework.

shaneutt:
Ready for re-review, added RAW urls and found explicit refs which match the versions.

joejulian (approved):

shaneutt:
Follow-up for future improvements (we need this now due to time constraints): [DCOS-59231](https://jira.mesosphere.com/browse/DCOS-59231)

alejandroEsc:
I dont agree with this as the logic for this can be part of kommander itself. I will not block though.

PR #223, opened 2019-09-26
Adding konvoyconfig

jr0d:


templates/konvoyconfig.yaml
```diff
@@ -0,0 +1,27 @@
...
+      enabled: true
+    - name: none
+      enabled: true
+  chartReference:
+    chart: konvoyconfig
+    repo: https://mesosphere.github.io/charts/staging
+    version: 0.0.2
+    values: |
+      ---
+      clusterHostname: ""
```
shaneutt: :thinking: would we want to add a small comment here explaining to the reader the full context of how and why this configuration option works like this?

shaneutt (approved):
One non-blocking comment (but still might be worth updating for).

Marking this as blocked on https://github.com/mesosphere/charts/pull/141 as that needs to merge before this can.

hectorj2f:
Yeah, let's wait until the PR is merged

templates/konvoyconfig.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - name: aws
+      enabled: true
+    - name: docker
+      enabled: true
+    - name: none
+      enabled: true
+  chartReference:
+    chart: konvoyconfig
+    repo: https://mesosphere.github.io/charts/staging
+    version: 0.0.2
```
hectorj2f: This version is 0.0.1

hectorj2f (changes requested):

jr0d:
@hectorj2f fixed

hectorj2f (approved):

PR #225, opened 2019-09-28
Upgrade prometheus-operator

gracedo:
Upgrade to v5.19.1

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -33,7 +33,7 @@ spec:
   chartReference:
     chart: prometheus-operator
     repo: https://mesosphere.github.io/charts/staging
-    version: 5.10.9
+    version: 5.19.1
     values: |
       ---
       defaultRules:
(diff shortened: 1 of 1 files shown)


shaneutt:
Looks like this is `ready`, but blocked on https://github.com/mesosphere/charts/pull/144

templates/prometheus.yaml
```diff
@@ -32,8 +32,8 @@ spec:
       enabled: true
   chartReference:
     chart: prometheus-operator
-    repo: https://mesosphere.github.io/charts/staging
-    version: 5.10.9
+    repo: https://gracedo.github.io/charts/staging
```
shaneutt: Let's just make sure we bump this back to mesosphere when we're unblocked before merge :+1:

shaneutt (approved):

hectorj2f:
@gracedo what is the state of this PR ?

s12chung (approved):

s12chung (changes requested):
ah, keep forgetting, please update the url in line 22, the url should point to the values of the chart. the current url is a bug pointing to the wrong chart and it should be updated when the values.yaml change on the chart.

shaneutt:
@s12chung I think we're just waiting on your re-review to :heavy_check_mark: and merge this one

s12chung (approved):

s12chung:
my bad. forgot about this one. ✅

PR #228, opened 2019-09-30
dex: bump version of kubeaddons-addon-initial and dex chart

hectorj2f:


timaa2k:
What's the reasoning behind this bump?

hectorj2f:
@timaa2k we have bumped the charts 1hour ago, those new versions have to be used here

PR #229, opened 2019-09-30
dex: requires cert-manager

hectorj2f:
We need to include cert-manager in the dependencies of dex, so it gets installed before.

### DIFF
--- a/templates/dex.yaml
+++ b/templates/dex.yaml
@@ -22,7 +22,6 @@ spec:
     matchLabels:
       kubeaddons.mesosphere.io/provides: ingresscontroller
       kubeaddons.mesosphere.io/name: cert-manager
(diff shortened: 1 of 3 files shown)


templates/dex.yaml
```diff
@@ -21,6 +21,8 @@ spec:
   requires:
     matchLabels:
       kubeaddons.mesosphere.io/provides: ingresscontroller
+      kubeaddons.mesosphere.io/name: cert-manager
+      values.chart.helm.kubeaddons.mesosphere.io/cert-manager-setup: "https://raw.githubusercontent.com/mesosphere/charts/6c43b8ab10108fb1adba5c6dd10e800e5f1abdd0/staging/cert-manager-setup/values.yaml"
```
joejulian: shouldn't need this line, the name label should be sufficient.

hectorj2f: I'll remove it

PR #233, opened 2019-10-01
stable/1.15.4: bump certmanager version

hectorj2f:


timaa2k:
Do we need to update the annotation at the top of the file because `values.yaml` has changed?

s12chung:
as @timaa2k mentioned, please change line 11 to match the correct default values for `commander`

s12chung:
thanks for the change :)

PR #236, opened 2019-10-02
Cherry-pick for 1.15.4 branch

jieyu:
Cherry-pick for 1.15.4 release branch

We should re-tag stable-1.15.4-2 release from this branch. The existing tag include an addon that we soon removed (kube-oidc-proxy) on master branch.

Also, kommander was bumped in master, which is not intended for 1.15.4-x releases. Otherwise, the link to /ops/portal/kommander will return `Cannot GET /kommander`.

### DIFF
--- a/templates/traefik-forward-auth.yaml
+++ b/templates/traefik-forward-auth.yaml
@@ -9,7 +9,6 @@ spec:
   requires:
     matchLabels:
       kubeaddons.mesosphere.io/name: dex
-      values.chart.helm.kubeaddons.mesosphere.io/traefik-forward-auth: "https://raw.githubusercontent.com/mesosphere/charts/6c43b8ab10108fb1adba5c6dd10e800e5f1abdd0/staging/traefik-forward-auth/values.yaml"
   chartReference:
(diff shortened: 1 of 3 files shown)


templates/traefik-forward-auth.yaml
```diff
@@ -9,7 +9,6 @@ spec:
   requires:
     matchLabels:
       kubeaddons.mesosphere.io/name: dex
-      values.chart.helm.kubeaddons.mesosphere.io/traefik-forward-auth: "https://raw.githubusercontent.com/mesosphere/charts/6c43b8ab10108fb1adba5c6dd10e800e5f1abdd0/staging/traefik-forward-auth/values.yaml"
```
s12chung: think we want to keep this for `kommander` work.

s12chung: resolving because we're simply trying to match history.

jieyu: yeah,  i just cherry-pick the change from here: https://github.com/mesosphere/kubeaddons-configs/pull/229/files

PR #246, opened 2019-10-03
[kommander] bump chart version and include KCL values.

alejandroEsc:


templates/kommander.yaml
```diff
@@ -10,13 +10,16 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    appversion.kubeaddons.mesosphere.io/kommander: "1.50.0"
+    appversion.kubeaddons.mesosphere.io/kommander: "1.114.5"
     endpoint.kubeaddons.mesosphere.io/kommander: /ops/portal/kommander
     values.chart.helm.kubeaddons.mesosphere.io/kommander: "https://raw.githubusercontent.com/mesosphere/charts/6c43b8ab10108fb1adba5c6dd10e800e5f1abdd0/stable/kommander/values.yaml"
```
s12chung: ah. need to update this too.

s12chung: we should really have CI check these.

alejandroEsc: yeah i am aware, not critical, but yeah. I hate really dislike that entry btw.

s12chung: think kubeadons needs to be able to get that metadata itself :p

PR #251, opened 2019-10-04
make kommander values link specific to commit

s12chung:
the existing URL changes over time since it points to a branch


shaneutt (approved):
We need to get rid of this hack, but :heavy_check_mark: for now good fix

PR #252, opened 2019-10-04
fix prometheus values url

s12chung:
it was pointing to the wrong chart...


shaneutt (approved):
:heavy_plus_sign::heavy_plus_sign:

PR #255, opened 2019-10-07
up memory requests and limits for kcl-cm

jaxesn:
@jimmidyson is there where you suggest we changing the limits/request?  I verified this change with the docker provisioner and the values take.  What do you think the request/limit should be?


jimmidyson:
Let's change it in the base chart instead now that we know the defaults are way too low.

PR #256, opened 2019-10-07
templates: update nvidia properties

hectorj2f:
I updated the nvidia properties accordingly to the new changes in the addons.

With 
```
name: aws
      enabled: false
```
we force users to set `true` whenever they want to run the `gpu`.

With this change https://github.com/mesosphere/kubeaddons-configs/compare/hectorj2f/enable_nvidia?expand=1#diff-48e90cb482008adf2d74147274088c73L19, the addon won't appear in the list of addons to be installed by default with these providers.

### DIFF
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -9,16 +9,13 @@ metadata:
     kubeaddons.mesosphere.io/provides: nvidia
   annotations:
     appversion.kubeaddons.mesosphere.io/nvidia: "0.1"
+    values.chart.helm.kubeaddons.mesosphere.io/nvidia: "https://raw.githubusercontent.com/mesosphere/charts/869a26a8d4061d5b407cd260eac2eee9149f2823/staging/nvidia/values.yaml"
 spec:
   kubernetes:
     minSupportedVersion: v1.15.0
   namespace: kube-system
   cloudProvider:
     - name: aws
-      enabled: false
-    - name: docker
-      enabled: false
-    - name: none
       enabled: false
   chartReference:
     chart: nvidia
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -17,6 +17,8 @@ spec:
   cloudProvider:
     - name: aws
       enabled: false
+    - name: none
+      enabled: false      
   chartReference:
     chart: nvidia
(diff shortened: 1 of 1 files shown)


templates/nvidia.yaml
```diff
@@ -9,17 +9,14 @@ metadata:
     kubeaddons.mesosphere.io/provides: nvidia
   annotations:
     appversion.kubeaddons.mesosphere.io/nvidia: "0.1"
+    values.chart.helm.kubeaddons.mesosphere.io/nvidia: "https://raw.githubusercontent.com/mesosphere/charts/869a26a8d4061d5b407cd260eac2eee9149f2823/staging/nvidia/values.yaml"
```
shaneutt: :+1:

templates/nvidia.yaml
```diff
@@ -9,17 +9,14 @@ metadata:
...
 spec:
   kubernetes:
     minSupportedVersion: v1.15.0
   namespace: kube-system
   cloudProvider:
     - name: aws
       enabled: false
-    - name: docker
-      enabled: false
-    - name: none
```
shaneutt: Just curious about the context: Why wont this work in on-prem? Could we potentially add some comments to this section so that the reader can at-a-glance quickly understand the reasons why we expect AWS only use for this addon?

shaneutt (changes requested):

templates/nvidia.yaml
hectorj2f: nope, you're right this makes sense in on-prem. I removed more than I should

shaneutt: ok cool, np!

templates/nvidia.yaml
```diff
@@ -9,17 +9,16 @@ metadata:
...
     minSupportedVersion: v1.15.0
   namespace: kube-system
   cloudProvider:
     - name: aws
       enabled: false
-    - name: docker
-      enabled: false
     - name: none
-      enabled: false
+      enabled: false
```
Gilbert88: Nits: remove additional spacing?

Gilbert88 (approved):
LGTM!

shaneutt (approved):

PR #268, opened 2019-10-09
prometheus: Set externalLabels cluster label to kube-system ns uid

gracedo:
Please see https://github.com/mesosphere/kubeaddons-extrasteps/pull/22 for a full description of this work. The extrasteps functionality is pulled into a prometheus container - ideally this would be an initContainer, but that feature was introduced in a later version of prometheus-operator ([0.32.0](https://github.com/coreos/prometheus-operator/releases/tag/v0.32.0), or helm chart versions >=6.8.0; we are on 5.19). When we eventually upgrade to a high enough version, we should move this into an initContainer. Until it is moved to an initContainer, there will be some errors (pasted below) in the reload and thanos sidecar containers until the configmap is written out and the env variable is defined, but it's pretty quick.

image
image
image

When the corresponding extrasteps and chart PRs are merged in, I'll update this to point to the correct chart/image.

### DIFF
--- a/templates/kibana.yaml
+++ b/templates/kibana.yaml
@@ -56,8 +56,8 @@ spec:
           - kibana-prometheus-exporter,6.7.0,https://github.com/pjhampton/kibana-prometheus-exporter/releases/download/6.7.0/kibana-prometheus-exporter-6.7.0.zip
       extraContainers: |
         - name: initialize-kibana-index
-          image: mesosphere/kubeaddons-addon-initializer:v0.0.4
-          command: ["/bin/bash", "-c", "addon-initializer kibana && while true; do sleep 86400; done"]
+          image: mesosphere/kubeaddons-addon-initializer:v0.0.13
+          command: ["/bin/bash", "-c", "addon-initializer kibana && sleep infinity"]
           env:
           - name: "KIBANA_NAMESPACE"
             value: "kubeaddons"
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -106,8 +106,8 @@ spec:
             # note: in order to run this container, mesosphereResources.enableAdditionalPrometheusRBACRules
             # needs to be set to true
             - name: get-cluster-id
-              image: mesosphere/kubeaddons-addon-initializer:v0.0.12
-              command: ["/bin/bash", "-c", "addon-initializer prometheus && while true; do sleep 86400; done"]
+              image: mesosphere/kubeaddons-addon-initializer:v0.0.13
+              command: ["/bin/bash", "-c", "addon-initializer prometheus && sleep infinity"]
               env:
               - name: "PROMETHEUS_NAMESPACE"
                 value: "kubeaddons"
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -19,7 +19,7 @@ metadata:
     docs.kubeaddons.mesosphere.io/prometheus: "https://prometheus.io/docs/introduction/overview/"
     docs.kubeaddons.mesosphere.io/grafana: "https://grafana.com/docs/"
     docs.kubeaddons.mesosphere.io/alertmanager: "https://prometheus.io/docs/alerting/alertmanager/"
-    values.chart.helm.kubeaddons.mesosphere.io/prometheus: "https://raw.githubusercontent.com/mesosphere/charts/6e22a56951dbe295c64f83b06f7c059fe4d612ce/staging/prometheus-operator/values.yaml"
+    values.chart.helm.kubeaddons.mesosphere.io/prometheus: "https://raw.githubusercontent.com/mesosphere/charts/3bc311e3fae9d608c528433963bf67ca6827513c/staging/prometheus-operator/values.yaml"
 spec:
   kubernetes:
     minSupportedVersion: v1.15.0
@@ -33,7 +33,7 @@ spec:
   chartReference:
     chart: prometheus-operator
     repo: https://mesosphere.github.io/charts/staging
-    version: 5.19.1
+    version: 5.19.2
     values: |
       ---
       defaultRules:
@@ -46,6 +46,10 @@ spec:
           # addon alert rules are defaulted to false to prevent potential misfires if addons
           # are disabled.
           velero: false
+        # This option grants the Prometheus pod the necessary permissions that the
+        # "get-cluster-id" container defined in prometheus.prometheusSpec.containers requires.
+        # see: https://github.com/mesosphere/kubeaddons-extrasteps/tree/master/pkg/prometheus/prometheus.go
+        enableAdditionalPrometheusRBACRules: true
       prometheus:
         service:
           additionalPorts:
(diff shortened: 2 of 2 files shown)


hectorj2f:
@gracedo will the thanos pod fail or just throw errors in the logs ?

gracedo:
@hectorj2f It'll just throw errors until the external labels get successfully set (see the last screencap)

gracedo:
@shaneutt per our discussion in slack yesterday, I'm going to make the change to add a values option for extra rbac rules in the chart instead, and set the rules here in kubeaddons-configs. I'll move these back into ready when those changes are in :)

shaneutt:
> @shaneutt per our discussion in slack yesterday, I'm going to make the change to add a values option for extra rbac rules in the chart instead, and set the rules here in kubeaddons-configs. I'll move these back into ready when those changes are in :)

Yes indeed, thank you for the reminder! :coffee: 

Let us know if you need any assistance with any of that! :vulcan_salute:

templates/prometheus.yaml
```diff
@@ -90,6 +108,23 @@ spec:
           thanos:
             baseImage: thanosio/thanos
             version: v0.4.0
+          externalLabels:
+            cluster: $(CLUSTER_ID)
+          containers:
+            - name: get-cluster-id
+              image: mesosphere/kubeaddons-addon-initializer:v0.0.12
+              command: ["/bin/bash", "-c", "addon-initializer prometheus && while true; do sleep 86400; done"]
```
joejulian:
```suggestion
              command: ["/bin/bash", "-c", "addon-initializer prometheus && sleep infinity"]
```

gracedo: seeing `sleep: invalid number 'infinity'` - doesn't look like `infinity` is supported, reverting

gracedo:
When https://github.com/mesosphere/kubeaddons-extrasteps/pull/23 is merged, we can change the sleep loop to `sleep infinity` (but doesn't need to block this PR, can open up a new PR with that change for all the areas where it's done) cc @joejulian

gracedo:
@joejulian @shaneutt @s12chung @hectorj2f I've pointed this back to the updated mesosphere chart, please review thank you! 🙏

templates/prometheus.yaml
joejulian: That took way longer than I'd hoped, but v0.0.13 is ready and fixes sleep.

gracedo: cool, thanks for doing that I'll make the change

templates/kibana.yaml
```diff
@@ -56,8 +56,8 @@ spec:
           - kibana-prometheus-exporter,6.7.0,https://github.com/pjhampton/kibana-prometheus-exporter/releases/download/6.7.0/kibana-prometheus-exporter-6.7.0.zip
       extraContainers: |
         - name: initialize-kibana-index
-          image: mesosphere/kubeaddons-addon-initializer:v0.0.4
-          command: ["/bin/bash", "-c", "addon-initializer kibana && while true; do sleep 86400; done"]
+          image: mesosphere/kubeaddons-addon-initializer:v0.0.13
+          command: ["/bin/bash", "-c", "addon-initializer kibana && sleep infinity"]
```
hectorj2f: @joejulian @shaneutt  this logic should be part of the add-on initializer passing a flag or something `--wait`. Also, could you remind me why we need to wait forever here?

templates/prometheus.yaml
```diff
@@ -90,6 +94,25 @@ spec:
           thanos:
             baseImage: thanosio/thanos
             version: v0.4.0
+          externalLabels:
+            cluster: $(CLUSTER_ID)
```
hectorj2f: Why do we need this pod sleeping forever?

templates/prometheus.yaml
```diff
@@ -90,6 +94,25 @@ spec:
...
+          containers:
+            # note: in order to run this container, mesosphereResources.enableAdditionalPrometheusRBACRules
+            # needs to be set to true
+            - name: get-cluster-id
+              image: mesosphere/kubeaddons-addon-initializer:v0.0.13
+              command: ["/bin/bash", "-c", "addon-initializer prometheus && sleep infinity"]
+              env:
+              - name: "PROMETHEUS_NAMESPACE"
+                value: "kubeaddons"
```
hectorj2f: What if we change the namespace? This should be more dynamic if we change the namespace name at any time

hectorj2f: Why cannot u get it as u do for the cluster_id?

templates/prometheus.yaml
```diff
@@ -90,6 +94,25 @@ spec:
...
+              - name: "CLUSTER_ID_CONFIGMAP_NAME"
+                value: "cluster-info-configmap"
+              - name: "KUBESYSTEM_NAMESPACE"
```
hectorj2f: The kube-system namespace will always be the same. This env variable seems redundant to me

templates/prometheus.yaml
gracedo: I think it's reasonable to have it static here - if we do change the namespace, this file would need to be updated in multiple locations since there are a bunch of references to the namespace (e.g. the metadata for this kubeaddon specifies `kubeaddons` ns statically)

templates/prometheus.yaml
gracedo: it's here as more of a precaution, but can remove. extrasteps defaults it to `metav1.NamespaceSystem` (`kube-system`) which is safe, but if we ever do for some reason need to change it, we can pass it in here to override then.

templates/prometheus.yaml
gracedo: Without the sleep, the container starts up then immediately exits after completing all the tasks, and causes a `CrashLoopBackOff`. IIUC once we update the prometheus-operator helm chart to a higher version, we can move this container into `initContainers` (not supported in current version) and avoid the sleep

templates/prometheus.yaml
gracedo: @joejulian I've made this change - mind re-reviewing?

templates/kibana.yaml
joejulian: Good point. I was only responding to the sleep loop but if this were an initContainer instead of a sidecar, exiting when finished would be ok. Is this possible, @gracedo?

https://github.com/helm/charts/blob/44d01d26c8b2b69066086b5788995680dfb6df40/stable/kibana/templates/deployment.yaml#L39-L46

gracedo: It does look like the version of Kibana we're using supports initContainers, so it should be able to be done (I guess we'll see if there was a reason why it wasn't done in the first place 🙃). However, this isn't technically related to this PR so I'm wary of making that change here. Would it be all right if we got this PR in first, and I'll make a follow-up PR for Kibana?

gracedo: https://jira.mesosphere.com/browse/DCOS-60303 for Kibana follow-up work

templates/prometheus.yaml
gracedo: https://jira.mesosphere.com/browse/DCOS-60304 for prometheus TODO once we upgrade to a version that supports `initContainers`

PR #270, opened 2019-10-11
default storage class addon and requirements

s12chung:
there has been a discussion about how to handle cases where there are 2 default storage classes, as that can create errors for other addons down the road.

part of this solution is having a validating webhook that checks for whether 2 default storage classes installed, see kep: https://github.com/mesosphere/kubeaddons/blob/master/keps/20190903-defaultstorageclass-validating-webhook.md

this is implemented in: https://github.com/mesosphere/charts/pull/124

to protect users from having multiple default storageclasses, all storageclass addons now have this addon as a requirement. as written in the PR.

### DIFF
--- a/templates/defaultstorageclass.yaml
+++ b/templates/defaultstorageclass.yaml
@@ -1,12 +1,12 @@
 apiVersion: kubeaddons.mesosphere.io/v1alpha1
 kind: Addon
 metadata:
-  name: defaultstorageclass
+  name: defaultstorageclass-protection
   namespace: kubeaddons
   labels:
-    kubeaddons.mesosphere.io/name: defaultstorageclass
+    kubeaddons.mesosphere.io/name: defaultstorageclass-protection
   annotations:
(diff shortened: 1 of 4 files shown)


shaneutt:
Hey @s12chung - sorry _not trying to block_ but could we fill out the description here with some context? It would be helpful to me to have a story here which describes the why (or if you can just link to JIRAs ans stuff that would be cool too!)

s12chung:
no worries. added.

templates/defaultstorageclass.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: kubeaddons.mesosphere.io/v1alpha1
+kind: Addon
+metadata:
+  name: defaultstorageclass
```
joejulian: I know it's a lame reason to block, but since this doesn't _provide_ the default storage class, can we pick a different name?

shaneutt: I wouldn't call that lame, it's legitimate to want it to be more easily understood (and less easily misunderstood) at a glance. :+1:

PR #272, opened 2019-10-11
gatekeeper: temporary disable it

hectorj2f:
I can reproduce 100% of the times the reset issue if I install gatekeeper.

### DIFF
--- a/templates/gatekeeper.yaml
+++ b/templates/gatekeeper.yaml
@@ -17,11 +17,11 @@ spec:
       kubeaddons.mesosphere.io/name: cert-manager
   cloudProvider:
     - name: aws
-      enabled: true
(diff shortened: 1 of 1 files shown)


joejulian:
> Let's investigate why this could cause reset issue next week.
Jira?

jieyu:
Ok, the theory here is that calico IPAM uses k8s as backend data store currently
https://github.com/mesosphere/konvoy/blob/master/ansible/roles/container-networking/templates/calico.yaml.tmpl#L719

If gatekeeper is installed, but shutdown during reset, the webhook service is gone (because the node it was running has been reset first). This would cause IPAM release IP to fail on other nodes when tearing down a container, causing reset to fail.

PR #273, opened 2019-10-11
[WIP] gatekeeper: re-enable it

hectorj2f:
We found that by deleting the gatekeeper webhook we manage to delete the pods without any problem, and so the reset operation works.


joejulian:
Since there are no changes from master, now, can this PR be closed?

PR #278, opened 2019-10-16
cert-manager: define konvoyconfig as dependency for cert-manager

hectorj2f:


hectorj2f:
We don't need this change.

PR #279, opened 2019-10-16
set cluster scoped label to addons

s12chung:
I'm sure about the first commit, not sure about the second.

related to: https://github.com/mesosphere/kubeaddons/pull/579


s12chung:
closing for now, as we're moving to `ClusterAddon`

PR #293, opened 2019-10-28
reloader: add new addon

mhrabovcin:
This PR adds new https://github.com/stakater/Reloader addon.

`Reloader` allows to explicitly set rules on which deployments will get a rolling update in a case when a config map or a secret contents get updated.

The proposed annotations configure `Traefik` to reload on `konvoconfig-kubeaddons` update and certificate change. When a user sets a new `clusterHostname` domain this will trigger the first reload of traefik. The init-container will update certificate specification and adds new domain. The `cert-manager` will pick up the change and it will re-issue the certificate. The secret that holds the certificate gets updated and that will trigger the second Traefik reload.

Other components are only reloaded when traefik certificate changes.

This PR also bumps multiple charts that will use this feature.

Depends on https://github.com/mesosphere/charts/pull/210

Konvoy PR that checks that tests have passed: https://github.com/mesosphere/konvoy/pull/944

### DIFF
--- a/templates/reloader.yaml
+++ b/templates/reloader.yaml
@@ -1,0 +1,29 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  name: reloader
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: reloader
+  annotations:
+    appversion.kubeaddons.mesosphere.io/dex: "v0.0.49"
+    values.chart.helm.kubeaddons.mesosphere.io/reloader: https://raw.githubusercontent.com/stakater/Reloader/ded923b12a17a0f5c9f37f1c4594331d8d22fa70/deployments/kubernetes/chart/reloader/values.yaml
+    # catalog.kubeaddons.mesosphere.io/addon-revision: "0.0.49-1"
+spec:
+  kubernetes:
+    minSupportedVersion: v1.15.0
+  chartReference:
+    chart: reloader
+    repo: https://stakater.github.io/stakater-charts
+    version: v0.0.49
+    values: |
+      ---
+      reloader:
+        deployment:
+          resources:
+            limits:
+              cpu: "100m"
+              memory: "512Mi"
+            requests:
+              cpu: "100m"
+              memory: "128Mi"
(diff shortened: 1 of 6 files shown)


templates/reloader.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: Addon
+metadata:
+  name: reloader
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: reloader
+  annotations:
+    appversion.kubeaddons.mesosphere.io/dex: "v0.0.49"
+    values.chart.helm.kubeaddons.mesosphere.io/reloader: https://raw.githubusercontent.com/stakater/Reloader/ded923b12a17a0f5c9f37f1c4594331d8d22fa70/deployments/kubernetes/chart/reloader/values.yaml
+    # catalog.kubeaddons.mesosphere.io/addon-revision: "0.0.49-1"
```
shaneutt: This has to be present to keep this repository compatible with the Kubeaddons Catalog:
```suggestion
    catalog.kubeaddons.mesosphere.io/addon-revision: "0.0.49-1"
```

templates/reloader.yaml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  kubernetes:
+    minSupportedVersion: v1.15.0
+  chartReference:
+    chart: reloader
+    repo: https://stakater.github.io/stakater-charts
```
shaneutt: Just curious if there's any particular context about this upstream, do you feel that there's any reason we may potentially may want to fork and bring this in house?

jaxesn:
Will/should this be enabled by default?

templates/reloader.yaml
mhrabovcin: So far we're using upstream chart without need to modify it.

mhrabovcin:
@jaxesn this controller will be enabled by default.

jaxesn:
I might just be missing it, but I though in one of these yaml files we had to explicitly say that's it enabled by default? Like per provider

samba:
Let's hold a minute here; unfortunately registering this as another default addon may have some unintended side-effects. Will discuss via Slack to assess impact & next steps.

mhrabovcin:
Adding `blocked` tag per @samba's comment.

mhrabovcin:
@hectorj2f we apply reloader configuration on `dex`, `dex-k8s-authenticator`, `kube-oidc-proxy`, `traefik-forward-auth` and `traefik` charts.

mhrabovcin:
After discussing in Slack with @jimmidyson  and @hectorj2f I am removing `blocked` label.

https://mesosphere.slack.com/archives/CMPPXKGKX/p1572341454399000

PR #294, opened 2019-10-28
integrating istio as an addon on konvoy

GoelDeepak:


templates/istio.yaml
```diff
@@ -0,0 +1,80 @@
...
+spec:
+  namespace: istio-system
+  requires:
+    - matchLabels:
+        kubeaddons.mesosphere.io/name: cert-manager
+  kubernetes:
+    minSupportedVersion: v1.15.0
+  cloudProvider:
+    - name: aws
+      enabled: false
```
shaneutt: Is it intentional for all the providers to be set to `enabled: false`?

GoelDeepak: yeah, the idea is to keep it `disabled` by default

templates/istio.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: istio
+  labels:
+    kubeaddons.mesosphere.io/name: istio
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.3.3-1"
+    appversion.kubeaddons.mesosphere.io/istio: "1.3.3"
+    endpoint.kubeaddons.mesosphere.io/kiali: "/ops/portal/kiali"
```
s12chung: I'm curious, is this a new annotation type?

GoelDeepak: I copied it from other YAMLs. I am under impression that it is used by the UI team but I could very well be wrong here :).

PR #296, opened 2019-10-30
templates: harbor deployment

hectorj2f:
Use `v1.3.0-kommander-beta3` release or latest master.


hectorj2f:
sure, reuse what you need

PR #308, opened 2019-11-04
Bump Nvidia chart version and make nvidia driver image tag configurable.

Gilbert88:
Bump the Nvidia helm chart to 0.1.2, which includes the image update for nvidia driver and nvidia device plugin.

Add the default nvidia driver image tag for centos, which make it configurable in Konvoy cluster.yaml.

### DIFF
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -7,9 +7,9 @@ metadata:
     kubeaddons.mesosphere.io/name: nvidia
     kubeaddons.mesosphere.io/provides: nvidia
   annotations:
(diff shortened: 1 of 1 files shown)


Gilbert88:
> There isn't a description. Can you explain why you disable it for the providers?

 @hectorj2f , Updated the description in PR. `cloudProvider` here means the nvidia addon default enabled option in Konvoy, right? We probably do not expect the addon is turned on by default.

> To me it seems the purpose of this is to select a specific image. If that's true ✔️ whenever Hector's comments are resolved.

@shaneutt , yes. This PR bumps the image versions, which is https://github.com/mesosphere/charts/pull/214

PR #326, opened 2019-11-15
templates: fix docker for the local volume provisioner

hectorj2f:
It seems this got merged but there was a wrong property set here: https://github.com/mesosphere/kubeaddons-configs/commit/0a5057a1f3dc7603ea3d40386fea88f149caa8d9#diff-940268ba9fc726ba58c83765edbbbf52


hectorj2f:
I am merging this PR so we can hopefully get the tests passing.

PR #328, opened 2019-11-15
dispatch: add initial kubeaddon

gpaul:
This currently points at my fork of the charts repo, please don't merge it yet.

### DIFF
--- a/templates/dispatch.yaml
+++ b/templates/dispatch.yaml
@@ -1,0 +1,39 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: dispatch
+  labels:
+    kubeaddons.mesosphere.io/name: dispatch
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.3.0-1"
+    appversion.kubeaddons.mesosphere.io/dispatch: "0.3.0"
+    endpoint.kubeaddons.mesosphere.io/dispatch: /dispatch/tekton/
+    docs.kubeaddons.mesosphere.io/dispatch: "https://beta-dispatch.d2iq.com/"
+    appversion.kubeaddons.mesosphere.io/argo-cd: "1.2.0"
+    endpoint.kubeaddons.mesosphere.io/argo-cd: /dispatch/argo-cd
+    docs.kubeaddons.mesosphere.io/argo-cd: "https://argoproj.github.io/argo-cd/user-guide/"
+    values.chart.helm.kubeaddons.mesosphere.io/dispatch: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/dispatch/values.yaml"
+spec:
+  namespace: dispatch
+  kubernetes:
+    minSupportedVersion: v1.15.0
+  cloudProvider:
+    - name: aws
+      enabled: false
+    - name: azure
+      enabled: false
+    - name: docker
+      enabled: false
+    - name: none
+      enabled: false
+  chartReference:
+    chart: dispatch
+    repo: https://mesosphere.github.io/charts/stable
+    version: 0.3.3
+    values: |
+      ---
+      global:
+        prometheus:
+          enabled: true
+          release: prometheus-kubeaddons


templates/dispatch.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: dispatch
+  labels:
+    kubeaddons.mesosphere.io/name: dispatch
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.3.0"
+    appversion.kubeaddons.mesosphere.io/dispatch: "0.3.0"
+    endpoint.kubeaddons.mesosphere.io/dispatch: /dispatch/tekton/
```
jbarrick-mesosphere: I was wondering if we should have this if we create our own Ingress records?

TattdCodeMonkey: This annotation is read by the konvoy/kommander UI and used to create the link for the punchout card.

If or when we have docs you would also want to add a link to those with:
`docs.kubeaddons.mesosphere.io/dispatch: "<URL_TO_DOCS>"`

templates/dispatch.yaml
```diff
@@ -0,0 +1,30 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: dispatch
+  labels:
+    kubeaddons.mesosphere.io/name: dispatch
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.3.0"
```
natmegs: It seems like most templates have `<appversion>-1` for revision but not 100% sure if the `-1` is required. Maybe @shaneutt can confirm what's expected here?

jbarrick-mesosphere:
> From the UI it appears we need to surface details about underlying components (argo, tekton, maybe others? anything with UI). Can we add appversion, endpoint, and docs annotations for each? Similar example is https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/prometheus.yaml

Thanks! Okay, I split argo-cd into its own add-on - and we'll call the Tekton UI "dispatch" for now, since it's more of an implementation detail than argo-cd is.

templates/dispatch.yaml
TattdCodeMonkey: this should be `/tekton:` so that the punch out is labels Tekton and not Dispatch.

also make sure the appversion matches the version you want to display in the UI for the link to Tekton.

natmegs: It sounds like we're calling tekton dispatch for now? https://github.com/mesosphere/kubeaddons-configs/pull/328#issuecomment-554516441

jbarrick-mesosphere: So I was thinking we would call it `dispatch` and `argo-cd` for now, since tekton is more of an implementation detail and eventually we'll have the Tekton UI fully wrapped (possibly deep link to it from the "Dispatch UI"). So I don't think we need a Tekton badge - we can use this as the entry point / logo / docs link for Dispatch.

TattdCodeMonkey:
if they are installed together they can be annotations on a single addon. 

we do this for prometheus, istio etc.

natmegs:
@TattdCodeMonkey the annotations for both are on this addon object

jbarrick-mesosphere:
:shipit: the status reports will have the wrong URL as we need to clean up our ingress handling, but that will be okay for the demo.

templates/dispatch.yaml
joejulian: Shane's out for the weekend. How do we test that this works in the frontend without the -1?

jbarrick-mesosphere: I tested this on my own konvoy cluster and everything installs. It does not show in the ops portal dashboard, but my understanding is it is not supposed to yet, according to @natmegs:

> if ingress is set up it should be reachable on the cluster but the links have not yet been added to kommander/konvoy UIs

natmegs: The part I'm wondering about is how the revision annotation could affect catalog API

shaneutt: The catalog will not support parsing this repo right now if this annotation isn't there, and in the correct format. It's a hack, we have tech debt JIRA stuff to clean this up but is necessary for now.

This should be changed to the following:

```suggestion
    catalog.kubeaddons.mesosphere.io/addon-revision: "0.3.0-1"
```

jbarrick-mesosphere:
> I dont know the details of this particular addon, however, i tried to deploy it and it works with the caveat that we must have 5 workers minimum for konvoy for it to deploy completely.

Yeah, I noticed this too, that's okay with me for now (it's not a default add-on). We need to do some work on our default requests since they're a bit greedy.

jbarrick-mesosphere:
Squashed and signed.

jbarrick-mesosphere:
> I'd like to confirm what's expected for the addon-revision annotation but the rest looks good

@natmegs do you know if there's anyone available who can do that? my understanding is this is blocking the konvoy release.

alejandroEsc:
commits dont seem to be signed?

jbarrick-mesosphere:
I ran `git commit -s --amend` but I guess I need a key?

jbarrick-mesosphere:
Fixed the signature.

PR #330, opened 2019-11-15
kommander: version bump and configure image version of kubeaddons

alejandroEsc:


templates/kommander.yaml
```diff
@@ -48,3 +48,10 @@ spec:
           deployment:
             annotations:
               configmap.reloader.stakater.com/reload: kommander-kubeaddons-config
+
+      federate:
+        kubeaddons:
+          image:
+            repository: "mesosphere/kubeaddons"
+            tag: "v0.5.3"
+            pullPolicy: IfNotPresent
```
s12chung: might be better to just use the default? non-blocking.

PR #340, opened 2019-11-21
chore: bump kommander/opsportal charts for image update 1.150.12

natmegs:
Depends on mesosphere/charts#278


alejandroEsc:
Tested out and works as expected!

PR #348, opened 2019-11-25
Update outdated appversion annotations

gracedo:
I went through the list of addons and updated some outdated appversion annotations.

### DIFF
--- a/templates/traefik.yaml
+++ b/templates/traefik.yaml
@@ -10,8 +10,8 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.68.4-2"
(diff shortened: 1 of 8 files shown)


templates/traefik.yaml
```diff
@@ -10,8 +10,8 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.68.4-2"
-    appversion.kubeaddons.mesosphere.io/traefik: "1.68.4"
```
gracedo: this appversion points to the chart version, is that right or should it be the version of traefik we're using? (`1.7.12`)

s12chung: should be the version of traefik.

gracedo: thx! fixed. I'll make a second pass tomorrow, I think I saw a couple others that referred to the chart instead of app version

PR #350, opened 2019-11-26
Bumped the Nvidia chart version to support GPU metrics.

Gilbert88:
This PR bumps the Nvidia chart version to 0.2.0, which contains the GPU metrics and granfana dashboard support.

It also removes the limits for GPU driver daemonset, because GPU is considered as dedicated resources and we do not want to add limited resources to those mission critical tasks like GPU driver, to avoid dependent applications being affected if the driver being OOM killed.

This PR depends on the helm chart PR https://github.com/mesosphere/charts/pull/285

Screen Shot 2019-11-26 at 3 27 32 PM
Screen Shot 2019-11-26 at 3 29 20 PM

### DIFF
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -53,6 +53,8 @@ spec:
           image: busybox
           command: ['sh', '-c', 'sleep 180']
       nvidia-dcgm-exporter:
+        enabled: true
+        namespace: kubeaddons
         nodeSelector:
           konvoy.mesosphere.com/gpu-provider: NVIDIA
         initContainers:
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -51,4 +51,4 @@ spec:
         initContainers:
         - name: init-wait
           image: busybox
-          command: ['sh', '-c', 'sleep 90']
+          command: ['sh', '-c', 'sleep 180']
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -52,3 +52,10 @@ spec:
         - name: init-wait
           image: busybox
           command: ['sh', '-c', 'sleep 180']
+      nvidia-dcgm-exporter:
+        nodeSelector:
+          konvoy.mesosphere.com/gpu-provider: NVIDIA
+        initContainers:
+        - name: init-wait
+          image: busybox
+          command: ['sh', '-c', 'sleep 200']
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -34,7 +34,7 @@ spec:
           tag: "418.87.01-centos7"
         resources:
           requests:
-             cpu: 200m
+             cpu: 500m
              memory: 512Mi
         nodeSelector:
           konvoy.mesosphere.com/gpu-provider: NVIDIA
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -61,3 +61,6 @@ spec:
         - name: init-wait
           image: busybox
           command: ['sh', '-c', 'sleep 200']
+      grafana:
+        enabled: true
+        namespace: kubeaddons
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -33,12 +33,9 @@ spec:
         image:
           tag: "418.87.01-centos7"
         resources:
-          limits:
-             cpu: 2000m
-             memory: 6144Mi
           requests:
              cpu: 200m
-             memory: 256Mi
+             memory: 512Mi
         nodeSelector:
           konvoy.mesosphere.com/gpu-provider: NVIDIA
       nvidia-device-plugin:
--- a/templates/nvidia.yaml
+++ b/templates/nvidia.yaml
@@ -7,8 +7,8 @@ metadata:
     kubeaddons.mesosphere.io/name: nvidia
     kubeaddons.mesosphere.io/provides: nvidia
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.1.1-1"
-    appversion.kubeaddons.mesosphere.io/nvidia: "0.1.1"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.2.0-1"
+    appversion.kubeaddons.mesosphere.io/nvidia: "0.2.0"
     values.chart.helm.kubeaddons.mesosphere.io/nvidia: "https://raw.githubusercontent.com/mesosphere/charts/master/staging/nvidia/values.yaml"
 spec:
   namespace: kube-system
@@ -26,7 +26,7 @@ spec:
   chartReference:
     chart: nvidia
     repo: https://mesosphere.github.io/charts/staging
-    version: 0.1.2
+    version: 0.2.0
     values: |
       ---
       nvidia-driver:
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -140,6 +140,22 @@ spec:
                   regex: '(.*):10250'
                   replacement: '${1}:1338'
                   target_label: __address__
+            - job_name: 'gpu_metrics'
+              scrape_interval: 1s
+              metrics_path: /gpu/metrics
+              tls_config:
+                ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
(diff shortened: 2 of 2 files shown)


alejandroEsc (dismissed):

templates/prometheus.yaml
```diff
@@ -140,6 +140,22 @@ spec:
                   regex: '(.*):10250'
                   replacement: '${1}:1338'
                   target_label: __address__
+            - job_name: 'gpu_metrics'
+              scrape_interval: 1s
```
branden: Is there a reason we need a shorter scrape interval here? The other targets use the default `1m` interval; `1s` is relatively quite frequent.

Gilbert88: Will `1s` cause a burden on Prometheus? 

I just use the `scrape_interval` defined from the official Nvidia monitoring tools:
https://github.com/NVIDIA/gpu-monitoring-tools/blob/master/exporters/prometheus-dcgm/k8s/pod-gpu-metrics-exporter/prometheus/prometheus-configmap.yaml#L18

I can relax it if necessary. Or we can tune it in the future when there is a perf impact.

shaneutt:
We need  to open this PR against https://github.com/mesosphere/kubernetes-base-addons as well please, as [we're in an interim phase and deprecating this repository](https://mesosphere.slack.com/archives/CQQ4HV5NY/p1575491975235000). In the next few days, we'll update this repo to point out that new PR's only need to be made against the new repo.

templates/prometheus.yaml
branden: I don't think a `1s` scrape interval will burden Prometheus. But there are a couple reasons to not set the interval here.

The first is that the scrape timeout can't exceed the scrape interval, and a `1s` timeout is pretty short. (The default timeout is `10s`.) So setting a short interval increases the likelihood that a scrape will be canceled if the target is slow to respond, leading to inconsistent collection and dropped metrics.

The other reason is that the Prometheus community recommends using the same scrape interval among different targets as much as possible. (Examples: [here](https://www.robustperception.io/keep-it-simple-scrape_interval-id), [here](https://github.com/prometheus/prometheus/issues/2353#issuecomment-273709709)) This makes configuration simpler, and also makes it easier to write queries. Queries that use functions like `rate()` need to take into account the sample rate of the underlying metric. This is easier to do if you can assume that all metrics are sampled at the same rate.

So if we don't have a compelling reason to set an interval here, let's use the default.

Gilbert88: Gotcha! Makes sense to me!

I will remove the 1s which would make it uses the default interval. Thanks, @branden !

Gilbert88:
> We need to open this PR against https://github.com/mesosphere/kubernetes-base-addons as well please, as [we're in an interim phase and deprecating this repository](https://mesosphere.slack.com/archives/CQQ4HV5NY/p1575491975235000). In the next few days, we'll update this repo to point out that new PR's only need to be made against the new repo.

Please see https://github.com/mesosphere/kubernetes-base-addons/pull/10

shaneutt (dismissed):
cool, ty! :heavy_check_mark:

branden (dismissed):

alejandroEsc (commented):
I do not know the details of the information here, but the overall changes to the addon look good. also with the prometheus changes there, if the nvidia addon is not enabled, prometheus should still work right?

alejandroEsc:
Please rebase.

templates/nvidia.yaml
```diff
@@ -26,19 +26,16 @@ spec:
...
     chart: nvidia
     repo: https://mesosphere.github.io/charts/staging
-    version: 0.1.2
+    version: 0.2.0
     values: |
       ---
       nvidia-driver:
         image:
           tag: "418.87.01-centos7"
         resources:
```
joejulian: @hectorj2f You've done a ton of work on resource parameters. Do you have any issue with this change?

templates/nvidia.yaml
```diff
@@ -54,4 +51,16 @@ spec:
         initContainers:
         - name: init-wait
           image: busybox
-          command: ['sh', '-c', 'sleep 90']
+          command: ['sh', '-c', 'sleep 180']
```
joejulian: Where do these sleeps come from? What are they waiting for?

It's generally preferable to make waits deterministic when possible.

templates/nvidia.yaml
```diff
@@ -54,4 +51,16 @@ spec:
...
+      nvidia-dcgm-exporter:
+        enabled: true
+        namespace: kubeaddons
+        nodeSelector:
+          konvoy.mesosphere.com/gpu-provider: NVIDIA
+        initContainers:
+        - name: init-wait
+          image: busybox
+          command: ['sh', '-c', 'sleep 200']
```
joejulian: Same comment regarding sleep as above.

joejulian (changes requested):

Gilbert88:
> I do not know the details of the information here, but the overall changes to the addon look good. also with the prometheus changes there, if the nvidia addon is not enabled, prometheus should still work right?

Correct. It just adds a Scrape in prometheus for exporting additional metrics for GPU. Prometheus should not be affacted whether or not the Nvidia is enbaled.

templates/nvidia.yaml
Gilbert88: Are you asking for why we remove the limits?

Because:
1. the driver can burst up to 8GB or more in the first 5 min. It is compiling kernel modules during runtime. Every time memory is throttled by the limits, the driver might be OOM killed.
2. GPU is expensive resources and we only run this driver on gpu machines. Without the driver up and running, there is no GPU jobs can be launched. We don't want to affect the GPU base driver due to memroy limits.

Basically, for mission critical applications, it should be ok to set its QoS as burstable without limits.

templates/nvidia.yaml
Gilbert88: I understand. However, the device plugin depends on the driver starts the server (which is after it compiles the kernel module and set up everything needed during runtime). There is supposed to be some endpoints to indicate the readiness of the driver in runtime but I haven't had time to figure it out yet and the driver is not open sourced.

We don't necessarily to add the sleep here - the device plugin will just keep crash looping until the driver starts to be functioning. However, the sleep here would avoid that noise.

templates/nvidia.yaml
Gilbert88: ditto

Gilbert88:
> Please rebase.
> 
> -- also we need to make same changes in the kubernetes-base-configs repo.

Rebased

shaneutt:
@Gilbert88 is this (and its [counterpart](https://github.com/mesosphere/kubernetes-base-addons/pull/10)) ready to change from `wip` to `ready`?

Gilbert88:
> @Gilbert88 is this (and its [counterpart](https://github.com/mesosphere/kubernetes-base-addons/pull/10)) ready to change from `wip` to `ready`?

It depends on the helm chart https://github.com/mesosphere/charts/pull/285

Should we change it to be ready once the https://github.com/mesosphere/charts/pull/285 got merged?

gracedo:
@Gilbert88 I think `ready` is appropriate here if you're done implementing and are ready for people to review - `blocked` is the label that denotes this PR is dependent on the charts PR getting merged and would be removed once it's merged! That's my understanding at least 😄

templates/nvidia.yaml
joejulian: Can we add a jira to come back at this. Waiting potentially multiples of 3 minutes to start a workload that depends on GPU would make me frustrated, as an operator.

Not a blocker for this PR.

joejulian (approved):

templates/nvidia.yaml
Gilbert88: https://jira.mesosphere.com/browse/DCOS-62148

It is orthogonal to the GPU workloads. GPU workloads can be spun up immediately. This is just the device plugin to sleep for 180 seconds when the Nvidia addon is installed. Even without this sleep, the plugin has to wait for the driver to be ready, which usually take 90 ~ 180 seconds (compile kernel modules during container runtime). So, the device plugin somehow has to wait.

shaneutt (approved):
:heavy_check_mark: once we're unblocked :+1:

hectorj2f (approved):
LGTM, let's see when we can get these improvements in https://jira.mesosphere.com/browse/DCOS-62148 ;)

joejulian (approved):

PR #353, opened 2019-12-04
chore: bump kommander/opsportal charts

natmegs:
⚠️ Depends on mesosphere/charts#293


alejandroEsc:
You may need to make this change in https://github.com/mesosphere/kubernetes-base-addons also

natmegs:
@alejandroEsc Thanks for the heads up! PR here https://github.com/mesosphere/kubernetes-base-addons/pull/3. Do we still need this one or can it be closed?

shaneutt:
both are needed for a couple days, soon we'll converge and then all new PR's should go to base addons.

PR #354, opened 2019-12-04
templates: addons adjust the resource limits

hectorj2f:
All these changes have been done by analyzing the resource usage on our soak cluster for this particular addons.


hectorj2f:
Screen Shot 2019-12-04 at 5 48 30 PM

hectorj2f:
we show some default green bars and stay away from warnings with the default settings.

PR #357, opened 2019-12-05
prometheus: kube-state-metrics

hectorj2f:
This PR is blocked by the chart PR with the respective changes

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -37,7 +37,7 @@ spec:
   chartReference:
     chart: prometheus-operator
     repo: https://mesosphere.github.io/charts/staging
-    version: 5.19.6
+    version: 5.19.7
     values: |
       ---
       defaultRules:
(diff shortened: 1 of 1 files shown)


shaneutt:
We need  to open this PR against https://github.com/mesosphere/kubernetes-base-addons as well please, as [we're in an interim phase and deprecating this repository](https://mesosphere.slack.com/archives/CQQ4HV5NY/p1575491975235000). In the next few days, we'll update this repo to point out that new PR's only need to be made against the new repo.

templates/prometheus.yaml
```diff
@@ -302,4 +302,4 @@ spec:
           # override the default k8s.gcr.io/kube-state-metrics repositry
           # containerd mirror functionality does not support pulling these images
           # TODO remove once https://github.com/containerd/containerd/issues/3756 is resolved
-          repository: gcr.io/google_containers/kube-state-metrics
+          repository: quay.io/coreos/kube-state-metrics
```
jimmidyson: Why has this changed? Sorry I don't have any context to review this PR so sorry for what could be a stupid question.

hectorj2f: that is the new image repository used by the official chart to use v1.8.0. That version isn't available for `gcr.io/google_containers/kube-state-metrics`

hectorj2f: I am being explicit here about this change

PR #368, opened 2019-12-12
Upgrade prometheus-operator

branden:
https://jira.mesosphere.com/browse/DCOS-61841
https://jira.mesosphere.com/browse/DCOS-60484

This depends on https://github.com/mesosphere/charts/pull/309.

Corresponding kubernetes-base-addons PR: https://github.com/mesosphere/kubernetes-base-addons/pull/20

## Testing

See the testing section in the description for https://github.com/mesosphere/charts/pull/309.

### DIFF
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -10,11 +10,11 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.31.1-1"
-    appversion.kubeaddons.mesosphere.io/prometheus-operator: "0.31.1"
-    appversion.kubeaddons.mesosphere.io/prometheus: "2.9.2"
-    appversion.kubeaddons.mesosphere.io/alertmanager: "0.17.0"
-    appversion.kubeaddons.mesosphere.io/grafana: "6.2.5"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.34.0-1"
+    appversion.kubeaddons.mesosphere.io/prometheus-operator: "0.34.0"
+    appversion.kubeaddons.mesosphere.io/prometheus: "2.14.0"
+    appversion.kubeaddons.mesosphere.io/alertmanager: "0.19.0"
+    appversion.kubeaddons.mesosphere.io/grafana: "6.4.2"
     endpoint.kubeaddons.mesosphere.io/prometheus: "/ops/portal/prometheus"
     endpoint.kubeaddons.mesosphere.io/alertmanager: "/ops/portal/alertmanager"
     endpoint.kubeaddons.mesosphere.io/grafana: "/ops/portal/grafana"
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -37,7 +37,7 @@ spec:
   chartReference:
     chart: prometheus-operator
     repo: https://mesosphere.github.io/charts/staging
-    version: 5.19.7
+    version: 8.3.3
     values: |
       ---
       defaultRules:
--- a/templates/prometheus.yaml
+++ b/templates/prometheus.yaml
@@ -107,6 +107,8 @@ spec:
         prometheusSpec:
           image:
             tag: v2.14.0
+          thanos:
+            version: v0.8.1
           externalLabels:
             cluster: $(CLUSTER_ID)
           containers:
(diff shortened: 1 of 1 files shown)


templates/prometheus.yaml
```diff
@@ -105,11 +105,6 @@ spec:
                 targetPort: 5601
                 interval: 30s
         prometheusSpec:
-          image:
-            tag: v2.9.2
```
gracedo: We pinned to an older version of Prometheus since we didn't like the new UI introduced past this version. We were hoping to upgrade to the new UI once [this enhancement](https://github.com/prometheus/prometheus/pull/5758) was merged which makes the UI more palatable - looks like it was released in v2.14.0, but the chart is on v2.13.1. What do you think about using [v2.14.0](https://github.com/prometheus/prometheus/releases/tag/v2.14.0)? It was released on 11/11, so it's about a month old. I try this version out and see if anything obvious breaks

templates/prometheus.yaml
```diff
@@ -105,11 +105,6 @@ spec:
...
-          thanos:
```
gracedo: We need the thanos block to run the sidecar

templates/prometheus.yaml
GoelDeepak: I would vote for us moving to v2.14.0 if not obvious breaks are found :)

branden: 👍 I'll set this to v2.14.0 and play with it.

templates/prometheus.yaml
branden: Thanks! I restored the `version`, but left the `baseImage` removed since it's the same value as the default: https://github.com/coreos/prometheus-operator/blob/8d44e0990230144177f97cf62ae4f43b1c4e3168/pkg/prometheus/statefulset_test.go#L39

gracedo:
Hmm.. I manually added the job that sets the grafana home dashboard, which worked. I wonder if for some reason this job isn't getting created/executed on install 🤔 I wasn't able to find any logs wrt the job which makes it harder to tell what's going on
image

branden:
I spent several hours trying to figure out why the home dashboard isn't getting configured, but I don't yet have an answer. Here are my observations so far:

* Running the job manually after the Prometheus addon is deployed works.
* Running `helm template` on the prometheus-operator chart (with the prometheus addon's chart values) shows the job resource in its output.
* Running `helm install mesosphere-staging/prometheus-operator --version=8.3.3 --name prometheus-kubeaddons-test --namespace kubeaddons -f prom-addon-values.yaml --debug --dry-run` shows the job resource in its output.
* Deploying a modified version of the prometheus-operator chart with the job's `hook-delete-policy` removed makes no apparent difference.
* Deploying Konvoy with the Prometheus addon disabled, and then manually deploying the prometheus-operator chart with the addon's chart values, fails with an error stating that one of the chart's resources already exists. I don't understand why this is happening.

My only conclusion right now is that the job is never created, rather than getting cleaned up after it runs (which is the normal behavior). I still don't know why the job isn't created when `helm template` and `helm install --dry-run` indicate that it should be. I also don't have a guess as to how upgrading the prometheus-operator chart caused this problem. And I don't now why I'm unable to deploy the prometheus-operator chart manually on a Konvoy cluster with the prometheus addon disabled.

I'll continue investigating tomorrow.

gracedo:
@s12chung @alejandroEsc I think you both have worked on this post-install hook to set the home dashboard in grafana - any ideas why this job may not be getting created?

branden:
This PR fixes a problem that's much worse than the one it introduces, so let's get this in and I'll follow up with another PR that fixes the home dashboard bug.

templates/prometheus.yaml
```diff
@@ -10,11 +10,11 @@ metadata:
...
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.31.1-1"
-    appversion.kubeaddons.mesosphere.io/prometheus-operator: "0.31.1"
-    appversion.kubeaddons.mesosphere.io/prometheus: "2.9.2"
-    appversion.kubeaddons.mesosphere.io/alertmanager: "0.17.0"
-    appversion.kubeaddons.mesosphere.io/grafana: "6.2.5"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.34.0-1"
+    appversion.kubeaddons.mesosphere.io/prometheus-operator: "0.34.0"
+    appversion.kubeaddons.mesosphere.io/prometheus: "2.14.0"
+    appversion.kubeaddons.mesosphere.io/alertmanager: "0.19.0"
+    appversion.kubeaddons.mesosphere.io/grafana: "6.4.2"
```
gracedo: should we also update the values url on https://github.com/mesosphere/kubeaddons-configs/pull/368/files#diff-848433d061871ec88294b7c52b46b08cR24 ?

PR #370, opened 2019-12-12
put dashboard in kubeaddons

jr0d:


jaxesn:
Will there be an issue with upgrading with it changing namespaces? If so, I'm fine if that isn't handled here initially just something we need to handle before 1.3 ga.

jr0d:
I am pretty sure helm will handle the upgrade case.. resources are all namespaced and the pod will get launched with the new argument.

shaneutt:
@jr0d in this interim before we have the new repository set up completely, please make a copy of this PR against https://github.com/mesosphere/kubernetes-base-addons, this wont be required for much longer once we fully deprecate this repo.

templates/kommander.yaml
```diff
@@ -29,8 +29,8 @@ spec:
       enabled: true
   chartReference:
     chart: kommander
-    repo: https://mesosphere.github.io/charts/stable
-    version: 0.2.27
+    repo: http://blacknode.net:8080/
```
hectorj2f: this needs to be changed before merging it

jr0d:
https://github.com/mesosphere/kubernetes-base-addons/pull/22

jr0d:
@shaneutt I've created the PR for kubernetes-base-addons

hectorj2f:
@jr0d did you have the chance to test it on a konvoy cluster ?

jr0d:
@hectorj2f sorry for the late reply, yes, this has been tested.

PR #371, opened 2019-12-16
[fluent-bit] update with audit log values.

alejandroEsc:
This was tested locally using an aws cluster.


jieyu:
You sure parser “docker” is correct given we use cri parser in the other section

alejandroEsc:
@jieyu i tested both options out and did not see an advantage/change over one or the other. We can change it to use CRI if you prefer.

PR #374, opened 2019-12-17
apply upgrade strategy

jr0d:
This is needed to support upgrading now that dashboard has migrated to the kubeaddons namespace


jr0d:
https://github.com/mesosphere/kubernetes-base-addons/pull/25

PR #375, opened 2019-12-17
chore: bump kommander/opsportal 1.171.10

juliangieseke:
depends on https://github.com/mesosphere/charts/pull/316

seems like this repo is still used? updating here as well. same question as in https://github.com/mesosphere/kubernetes-base-addons/pull/26 : that opsportal link pointed to a specific git sha and was therefore never actually updated 🤔 bug?

### DIFF
--- a/templates/opsportal.yaml
+++ b/templates/opsportal.yaml
@@ -26,7 +26,7 @@ spec:
   chartReference:
     chart: opsportal
     repo: https://mesosphere.github.io/charts/stable
-    version: 0.1.22
+    version: 0.1.23
     values: |
       ---
       landing:
--- a/templates/kommander.yaml
+++ b/templates/kommander.yaml
@@ -6,8 +6,8 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: kommander
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.169.3-1"
-    appversion.kubeaddons.mesosphere.io/kommander: "1.169.3"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.171.10-1"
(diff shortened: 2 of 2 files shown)


templates/opsportal.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     catalog.kubeaddons.mesosphere.io/addon-revision: "1.0.0-1"
     appversion.kubeaddons.mesosphere.io/opsportal: "1.0.0"
     endpoint.kubeaddons.mesosphere.io/opsportal: /ops/portal/
-    values.chart.helm.kubeaddons.mesosphere.io/opsportal: "https://raw.githubusercontent.com/mesosphere/charts/4155f480571eaf82c64ddd63d3d334b1105d0591/stable/opsportal/values.yaml"
+    values.chart.helm.kubeaddons.mesosphere.io/opsportal: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/opsportal/values.yaml"
```
juliangieseke: this line ^

shaneutt (dismissed):
This repo is still used. The plan is to _fully deprecate_ and put up notice of it in this repo later today, as recent updates to Konvoy have allowed us to finally fully deprecate it.

templates/opsportal.yaml
natmegs: I don't think we want this pointing to master... I've never changed this value. @shaneutt thoughts?

juliangieseke: no, shane explained in new repo, we have to update that sha after charts repo PR has merged 🤷‍♂

shaneutt (approved):

juliangieseke:
opening new one…

